PR #474, opened 2026-03-16
feat: Integrate Qwen vLLM with FluidMCP

zeeshana-fluidai:
- Added start.py to run FluidMCP + Qwen vLLM as a single server
- Added cli_patched.py with None check fix for non-registry MCP packages
- Added qwen_mcp_server.py to proxy Qwen vLLM endpoints
- Updated config.json with install_path for filesystem and memory servers
- Added example config for Qwen vLLM integration

### DIFF
--- a/start.py
+++ b/start.py
@@ -1,0 +1,57 @@
+import json, os
+from pathlib import Path
+from fluidai_mcp.cli import extract_config_from_file
+from fluidai_mcp.services.package_launcher import launch_mcp_using_fastapi_proxy
+from fastapi import FastAPI, Request, APIRouter, Body
+from fastapi.responses import JSONResponse
+from pydantic import BaseModel
+from typing import List, Optional
+import requests as req_lib
+import uvicorn
+
+config = extract_config_from_file('/app/config.json')
+
+app = FastAPI(title="FluidMCP + Qwen vLLM", description="FluidMCP with Qwen vLLM integrated")
+
+# Pydantic Models
+class Message(BaseModel):
+    role: str
+    content: str
+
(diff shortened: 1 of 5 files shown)


start.py
```diff
@@ -0,0 +1,57 @@
...
+# MCP Servers
+for server_name, server_config in config["mcpServers"].items():
+    install_path = Path(server_config["install_path"])
+    print(f"Launching {server_name}...")
+    pkg_name, router = launch_mcp_using_fastapi_proxy(install_path)
+    app.include_router(router)
+    print(f"✅ {server_name} ready")
+
+print("🚀 Starting FluidMCP server...")
+uvicorn.run(app, host="0.0.0.0", port=8099)
```
abhinav-fluidai, line 57: This exposes the combined Qwen proxy and MCP routers on all interfaces, but the new integration registers those routes without any auth or authorization check. With the filesystem and memory MCP servers configured, any network client that reaches port 8099 can invoke model and MCP operations. Please gate the app with the existing secure-mode bearer enforcement or another required auth layer before binding it publicly.

abhinav-fluidai (changes requested):
Security review found a blocking exposure in the new FluidMCP + Qwen gateway. The PR registers the Qwen proxy and MCP routers, then serves the combined FastAPI app publicly without an authentication or authorization layer. Because the configured MCP servers include filesystem and memory, anyone who can reach the box on port 8099 could invoke model and MCP operations. Please require the existing secure-mode bearer enforcement or an equivalent auth layer before exposing these routes. Readiness discussion had no prior unresolved GitHub or Linear blockers, and no large-PR split is needed.

PR #515, opened 2026-03-26
[MCP-413]-fix: Production stability — stderr drainer, readline timeouts, crash …

srikar-reddy-fluidai:
# fix: Production stability — stderr drainer, readline timeouts, crash watchdog, lock race fixes

## What was broken

FluidMCP was going down in production without clear explanation. The root causes were a combination of silent process hangs, blocking I/O in the async event loop, undetected crashes, and a concurrency race condition that could leave servers in a permanently locked state.

---

### stderr pipe buffer deadlock (silent hang)

When nobody reads a subprocess's stderr pipe, the OS pipe buffer fills up (64KB limit) and the subprocess freezes waiting for the buffer to drain. This caused MCP servers to silently stop responding with no error logged.

**Fix:** Added a `_start_stderr_drainer` — a daemon thread that continuously reads and buffers stderr for every subprocess. The buffer holds the last 200 lines in memory. This prevents the pipe from ever filling up.

---

### `readline()` blocking the event loop indefinitely

All subprocess stdout reads used bare `process.stdout.readline()` with no timeout. If a process hangs or produces no output, this blocks forever — freezing the entire asyncio event loop and making the whole server unresponsive.

**Fix:** Replaced every `readline()` call with `_readline_with_timeout()`, which uses `select()` to wait with a 30-second timeout. If no data arrives, it logs a warning and returns an empty string instead of blocking.

---

### Crashes going undetected

When an MCP server process crashed, the server state in the database was never updated. The gateway kept routing requests to a dead process, and there were no crash logs to diagnose what went wrong.

**Fix:** Added `_watch_process` — an async watchdog task spawned for each server after startup. It waits for the process to exit, captures the last 50 lines of stderr as a crash log, saves it to the database, updates the server state, and triggers auto-restart if configured.

---

### Lock race condition (TOCTOU bug)

The operation locks used `lock.locked()` to check before acquiring — a classic time-of-check/time-of-use race. Two concurrent operations could both read `locked() == False` and both proceed, potentially double-starting or double-stopping a server.

**Fix:** Replaced the two-step check-then-acquire with `asyncio.wait_for(lock.acquire(), timeout=0)`. This is atomic — it either acquires immediately or raises `TimeoutError`. No race window.

---

### Watchdog triggering auto-restart on intentional stop

The watchdog fires when a process exits. On manual stop, the process is killed intentionally — but the watchdog would still detect the exit and try to restart the server.

**Fix:** The watchdog task is cancelled before sending SIGTERM during a stop operation. The watchdog only triggers if the process exits while it's still running — i.e., a real crash.

---

### `process.wait()` after SIGKILL had no timeout

After sending SIGKILL, the code called `process.wait()` with no timeout. If the OS was slow to reap the process, this could block indefinitely during a shutdown.

**Fix:** Added a 5-second timeout to `process.wait()` after SIGKILL.

---

### `stderr.read()` blocking in SSE handshake

The SSE subprocess handshake read the full stderr with `process.stderr.read()` — a blocking call that reads until EOF. Since the process is still running, this would never return.

**Fix:** Replaced with a non-blocking call to `get_stderr_tail()` wrapped in `asyncio.to_thread` with a timeout.

---

### Infinite log retry loop

The database log retry task had no upper bound on retries. If the database was permanently down, the retry task would loop indefinitely, consuming resources and filling logs.

**Fix:** Capped retries at 10 attempts with exponential backoff (up to 5 minutes between attempts). After the cap, buffered logs are dropped and a warning is logged.

---

### Broken import in `__main__.py`

The package entry point had `from cli import main` — a relative import written as absolute, which broke after a monorepo restructure. This caused `python -m fluidmcp.cli` to fail with an `ImportError`.

**Fix:** Changed to `from .cli import main`.

---

### 3-value unpack crash in `cli.py`

`launch_mcp_using_fastapi_proxy` was updated to return a 3-tuple `(package_name, router, process)` but the caller in `cli.py` still unpacked 2 values. This caused an immediate `ValueError` when using the `github` command.

**Fix:** Updated the unpack to `package_name, router, _process = launch_mcp_using_fastapi_proxy(...)`.

---

### E2E test false failure

The E2E test `test_github_clone_launch_and_jsonrpc_request` was failing due to `uv 0.7.2` being installed, but the MCP Python SDK requires `uv >= 0.9.5`. This was a pre-existing environment issue unrelated to FluidMCP code.

**Fix:** Added a `uv` version check at the top of the test — the test is skipped with a clear message if the installed `uv` version is too old.

---
## How to Test

pytest tests/test_llm_security.py tests/test_llm_integration.py tests/test_replicate_client.py

## Result

All unit tests pass. The production execution path (`fmcp serve`) now has:
- No blocking I/O in the event loop
- Automatic crash detection and logging
- Clean process lifecycle management
- No silent hangs from stderr buffer overflow


shubhendu-shekhar-fluidai (approved):

abhinav-fluidai (approved):
## Security Review

**Blast Radius: HIGH**
Touches the core MCP server lifecycle: `server_manager.py`, `package_launcher.py`, `management.py`, and `run_servers.py`. Wide surface, but all changes are narrowly scoped bug fixes.

### Summary
Production stability PR fixing stderr deadlocks, blocking readline in the async event loop, undetected subprocess crashes, and TOCTOU lock races. No auth changes, no new external surface, no secrets. The `httpx` timeout change (None→120s) and bounded log retry (10 attempts with exponential backoff) are positive security improvements.

### Issues Found
None.

### Verdict
APPROVE

abhinav-fluidai (approved):
## Security Review

**Blast Radius: HIGH**
Touches `server_manager.py`, `package_launcher.py`, `management.py` — core MCP process lifecycle infrastructure.

### Summary
Major stability overhaul: stderr drain threads (bounded deque, maxlen=200), unified `readline_with_timeout()` with `select()`/thread fallback for Windows, crash watchdog with exponential backoff restart, TOCTOU lock fix in `start/stop/restart_server`. Comprehensive — replaces several one-off timeout hacks with a consistent pattern.

### Issues Found
None.

### Verdict
APPROVE

PR #558, opened 2026-04-03
MCP-443 feat: add per-server async queue to serialize MCP stdio I/O

dhanesh-gore-fluidai:
## Problem

In `fmcp serve` mode, multiple concurrent HTTP requests share the same MCP subprocess stdin/stdout pipes with no serialization. The `create_dynamic_router()` endpoints directly called `process.stdin.write` + `asyncio.to_thread(stdout.readline)` without any request queuing. Under concurrent load this causes:

- Mixed/swapped responses between requests
- Race conditions on the pipe
- System instability

## Solution

Introduce a `McpQueue` per stdio subprocess that funnels all pipe I/O through a single background drain loop, serializing access completely.

### New: `fluidmcp/cli/services/mcp_queue.py`
- `McpQueue(server_name, process)` with a single `_drain_loop` asyncio task
- `async send(msg, timeout)` — enqueues request, returns response; uses `asyncio.shield(future)` so client disconnects don't corrupt pipe state
- `async write_stdin_only(msg)` — for SSE multi-line path (must hold `io_lock`)
- Public `io_lock: asyncio.Lock` — shared serialization point for both the drain loop and the SSE streaming path
- Hard 30s timeout on `stdout.readline()` inside drain loop to prevent hangs
- Queue size configurable via `FMCP_QUEUE_SIZE` env var (default: 100)
- `SSE_MAX_LINES = 500` safety cap for streaming loops

### Modified: `fluidmcp/cli/services/server_manager.py`
- `self.queues: Dict[str, McpQueue]` added to track per-server queues
- Queue started after process registration in `_start_server_unlocked` (skipped for `SseSubprocessHandle`)
- Queue stopped before process termination in `_stop_server_unlocked`
- Safety cleanup in `_cleanup_server` for processes that die on their own (crash/self-exit)
- New `get_queue(server_id) -> Optional[McpQueue]` accessor

### Modified: `fluidmcp/cli/services/package_launcher.py` (`create_dynamic_router` only)

| Endpoint | Before | After |
|---|---|---|
| `POST /{server}/mcp` | direct stdin/stdout | `queue.send(timeout=30)` |
| `POST /{server}/sse` | direct stdin/stdout loop | `async with queue.io_lock` + `write_stdin_only` + 500-line cap |
| `GET /{server}/mcp/tools/list` | direct stdin/stdout | `queue.send(timeout=30)` |
| `POST /{server}/mcp/tools/call` | direct stdin/stdout | `queue.send(timeout=60)` |

Error mapping: `QueueFull → 429`, `TimeoutError → 504`, `queue is None → 503`

## Scope

Only `fmcp serve` dynamic dispatch path is touched. `run_servers.py`, management API, auth, database, and frontend are untouched.

## Test plan
- [x] Run existing tests: `pytest tests/test_llm_security.py tests/test_llm_integration.py -v`
- [x] Start server: `fluidmcp run examples/sample-config.json --file --start-server`
- [x] Fire 10 concurrent requests and verify each response `id` matches the request `id` (no mixing)
- [x] Verify HTTP 429 is returned when queue is saturated
- [x] Verify HTTP 504 is returned on slow/hung MCP server
- [x] Verify SSE streaming still works end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


fluidmcp/cli/services/mcp_queue.py
```diff
@@ -0,0 +1,297 @@
+"""
+McpQueue — per-server async queue for serializing MCP subprocess stdio I/O.
+
+All writes to stdin and reads from stdout for a given subprocess are
+funnelled through a single drain loop coroutine, eliminating the race
+condition that occurs when multiple concurrent HTTP requests attempt to
```
rohan-ghorpade-fluidai: @dhanesh-gore-fluidai The codebase already has _get_io_lock(server_name) -- an asyncio.Lock per server that serializes stdin write + stdout read atomically. No response mixing can happen under the lock.

fluidmcp/cli/services/mcp_queue.py
```diff
@@ -0,0 +1,297 @@
...
+_DRAIN_READ_TIMEOUT = 30.0
+
+# Maximum number of response lines the SSE streaming path will read in a
+# single lock-hold before giving up, to prevent an infinite loop if the
+# subprocess never emits a "result" field.
+_SSE_MAX_LINES = 500
+
+
+class McpQueue:
+    """
```
rohan-ghorpade-fluidai: @dhanesh-gore-fluidai STDIO is inherently serial (one request at a time)

rohan-ghorpade-fluidai (changes requested):

fluidmcp/cli/services/server_manager.py
```diff
@@ -1119,14 +1130,8 @@ async def _cleanup_server(self, id: str, exit_code: int, intentional: bool = Fal
...
 
-        # Close stderr log file handle
-        self._close_stderr_log(id)
-
-        # Determine state
-        if intentional or exit_code == 0:
-            state = "stopped"
-        else:
-            state = "failed"
+        # Remove I/O lock for servers that died on their own (crash/self-exit).
```
abhinav-fluidai: This deletion removes the only assignment to `state`, but `_cleanup_server` still passes `state` to `save_instance_state` below. Every cleanup now raises `NameError`, preventing reliable lifecycle persistence; it also drops `_close_stderr_log(id)`. Restore the stopped/failed state calculation and stderr-log cleanup while retaining the I/O-lock removal.

fluidmcp/cli/services/package_launcher.py
```diff
@@ -763,48 +770,63 @@ async def event_generator() -> AsyncIterator[str]:
...
+                        # Record in global error metric for monitoring
+                        collector.record_error("io_error")
+                        yield f"data: {json.dumps({'error': f'Process pipe broken: {str(e)}'})}\n\n"
+                        return
+
+                    lines_read = 0
+                    while lines_read < _SSE_MAX_LINES:
+                        # Non-blocking I/O with asyncio.to_thread
+                        response_line = await asyncio.to_thread(process.stdout.readline)
+                        if not response_line:
```
abhinav-fluidai: `asyncio.to_thread()` cancellation does not stop the underlying blocking `stdout.readline`. If an SSE client disconnects while this read is pending, cancellation exits `async with io_lock` and releases the lock while the worker thread continues reading; the next request can then start a second reader on the same stdout, recreating response theft/corruption and enabling request-driven service disruption. Use a cancellation-safe single-owner reader/queue design (or otherwise ensure the background read has finished before releasing ownership), with a bounded timeout.

abhinav-fluidai (changes requested):
Readiness is blocked because GitHub reports merge conflicts. Please resolve the conflicts and address the inline cleanup regression and cancellation-safe stdio serialization issue before re-review. Security blast radius is HIGH because this changes shared request handling for every stdio-backed MCP server.

fluidmcp/cli/services/server_manager.py
dhanesh-gore-fluidai: Checked — this is already resolved at the current HEAD. The merge of `main` into this branch (commit 6dcc180) brought back both the `state` calculation and the `_close_stderr_log(id)` call right after the I/O-lock removal, so `_cleanup_server` no longer raises `NameError`. Ran `tests/test_server_manager_cleanup.py` against current HEAD — all 37 tests pass. No code change needed here.

fluidmcp/cli/services/package_launcher.py
dhanesh-gore-fluidai: Fixed in 0d51412. Digging into this, the merge of `main` into this branch (6dcc180) actually mangled `create_dynamic_router` far worse than this comment describes — the file had an `IndentationError` and didn't even compile, `proxy_jsonrpc` did a duplicate unlocked stdout read (every `/mcp` call would hang ~30s then 504), and `list_tools`/`call_tool` never wrote the request to stdin at all despite fetching the I/O lock (every call would time out). All four handlers are restored to write-then-read-under-lock.

For the cancellation-safety point specifically: added `_read_response_locked()`, which wraps the `asyncio.to_thread(readline_with_timeout, ...)` call in `asyncio.shield()`. If the awaiting coroutine is cancelled (client disconnect), it catches the `CancelledError`, awaits the shielded task to actually finish (bounded to the existing 30s/60s `readline_with_timeout` timeout), and only then re-raises — so `async with io_lock` doesn't release until the abandoned reader has actually stopped touching the pipe. Applied consistently to all four stdio handlers (`/mcp`, `/sse`, `/mcp/tools/list`, `/mcp/tools/call`), not just SSE.

Verified: module imports cleanly, `tests/test_llm_security.py` + `tests/test_llm_integration.py` + `tests/test_server_manager_cleanup.py` all pass (73 tests), and I ran a smoke test hitting all three JSON endpoints against a fake subprocess to confirm they now correctly write to stdin and return real responses.

dhanesh-gore-fluidai:
Addressed the readiness-gate blocker (\`package_launcher.py:1011\`) in fb31ac7.

**Issue:** the SSE timeout branch (\`continue\` on an empty/timed-out read) never advanced \`lines_read\`, so a subprocess that stays alive but never answers could hold the loop — and the shared per-server \`io_lock\` — open forever, starving every other request to that server.

**Fix:** added a bounded consecutive-timeout counter, \`_SSE_MAX_CONSECUTIVE_TIMEOUTS\` (default 20, i.e. ~10 min of total silence at the existing 30s per-read timeout; overridable via \`FMCP_SSE_MAX_CONSECUTIVE_TIMEOUTS\`). Once the cap is hit the stream logs a warning, yields an idle-timeout error event, and breaks out of the loop — releasing \`io_lock\` so other requests to that server aren't starved.

Verified with a smoke test simulating a subprocess that's alive but never responds: the stream now reliably terminates after exactly the configured number of timeouts, and a second concurrent request to the same server succeeds immediately afterward (confirming the lock is actually released, not held).

On the "pending GitHub review" 422 — that's an existing unsubmitted draft review on the reviewing account's side, not something addressable from this branch. Whoever owns that automation/account will need to submit or dismiss it before the next review pass.

PR #613, opened 2026-04-14
MCP Inspector 4A: Widget sandbox scroll & layout fixes

shubhendu-shekhar-fluidai:
## Summary
- **Fix chat scroll completely broken** -- the scroll container was also a flex container, causing children to shrink-to-fit instead of overflowing. Separated into outer scroll div + inner flex wrapper.
- **Cap result bubble at 350px** with internal scroll (WhatsApp-style cards) so the widget below is always reachable without scrolling through hundreds of lines of JSON.
- **Move auto-scroll anchor after widget** so `scrollIntoView` reveals the full widget, not just the result above it.
- **Override `.dashboard` CSS class** constraints (`max-width: 1200px`, `padding: 32px`) that clipped inspector content in normal mode.
- **Increase WidgetSandbox default height** from 280px to 400px and clean up redundant double-border wrapper.
- **Remove double maxHeight** constraint from ToolResult manual mode.

## Test plan
- [ ] Connect to a weather-ui MCP server (or any server with `_meta["ui/resourceUri"]`)
- [ ] In Chat mode, send a query that triggers a tool with a widget (e.g. "weather in Delhi")
- [ ] Verify: result card is capped at ~350px with its own scrollbar for large JSON
- [ ] Verify: widget renders fully below the result and is scrolled into view automatically
- [ ] Verify: in normal (non-fullscreen) mode, content is not clipped at the bottom
- [ ] Verify: Manual mode results still scroll properly
- [ ] Test in fullscreen mode -- same scroll behavior should work

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/fluidmcp/cli/.env.example
+++ b/fluidmcp/cli/.env.example
@@ -15,7 +15,8 @@
 # FMCP_GITHUB_TOKEN=ghp_yourTokenHere
 # GITHUB_TOKEN=ghp_yourTokenHere
 
-# Inspector is using Groq's model, so GROQ_API_KEY is required for InspectorAgent to function
+# Inspector chat uses Groq's model; GROQ_API_KEY is required only for InspectorAgent chat features,
+# not for basic inspector connectivity
 #GROQ_API_KEY=gsk_yourTokenHere 
 
 # S3 Credentials - REQUIRED only when using --master mode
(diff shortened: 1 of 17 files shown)


shubhendu-shekhar-fluidai:
Superseded — splitting into two smaller PRs (3C + 4A) rebased on #604

PR #616, opened 2026-04-14
MCP-475: MCP Inspector 3C: Chat revamp, log filters, UI polish

shubhendu-shekhar-fluidai:
## Summary
Stacks on #604 (widget sandbox).

- Per-server chat memory (`chatHistoryByServer`) and logs (`logsByServer`) — switching servers preserves history
- Log filter pills `[All][Connect][Tool][Error][Chat]` with live counts and color-coded borders
- System prompt dialog and chat toolbar (model badge, clear chat button)
- Right-aligned user message bubbles, cleaned key→value tool param display
- Active server indigo left-border accent
- Fullscreen mode — hides Navbar, maximises inspector to full viewport
- Widget detection via `_meta["ui/resourceUri"]` on tool results
- `ChatResultBubble`: text preview on collapse, `hideTabSwitcher` prop, word-break fix

**1 file changed, ~270 insertions, ~170 deletions** (MCPInspector.tsx only)

## Test plan
- [ ] Connect to an MCP server, switch between servers — chat and logs persist per server
- [ ] Log filter pills show correct counts and filter correctly
- [ ] System prompt dialog works, prompt sent with chat messages
- [ ] User bubbles right-aligned, tool params display cleanly
- [ ] Fullscreen button — Navbar hides, inspector fills viewport
- [ ] Server with `_meta["ui/resourceUri"]` tools — widget renders below result

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #617, opened 2026-04-14
MCP-480: MCP Inspector 4A: Widget sandbox scroll and layout fixes

shubhendu-shekhar-fluidai:
### Summary
Stacks on #616 (3C chat polish), which stacks on #604 (widget sandbox).

**Scroll + Layout Fixes**

- Fix chat scroll completely broken — scroll container was also a flex container, causing children to shrink-to-fit instead of overflowing. Separated into outer scroll div + inner flex wrapper
- Cap result bubble at 350px with internal scroll (WhatsApp-style cards) so widget below is always reachable
- Move auto-scroll anchor after widget so scrollIntoView reveals the full widget
- Override .dashboard CSS constraints (max-width: 1200px, padding: 32px) that clipped content in normal mode
- Increase WidgetSandbox default height 280px → 400px, clean up redundant double-border wrapper
- Remove double maxHeight constraint from ToolResult manual mode

**Widget Fullscreen**

- Add fullscreen toggle button (⤢/✕) in widget header bar — overlays entire viewport at position: fixed; inset: 0; z-index: 9999
- Single stable DOM tree — normal and fullscreen modes share the same React tree so the iframe is never remounted; widget state (game progress, app state) survives the toggle
- Widget owns its own header — removed duplicate external header from MCPInspector, WidgetSandbox renders the header+button itself

Bug Fixes

- Fix "Widget resource returned empty HTML" on HTTP (/mcp) transport — correctly unwraps MCP spec response shape {contents: [{uri, mimeType, text}]} instead of expecting a top-level text field
- tsconfig.app.json — remove deprecated baseUrl (redundant with moduleResolution: bundler), clears IDE warnings

**5 files changed across 2 commits**

**Test plan**

- [x]  Chat mode: large JSON result caps at ~350px with its own scrollbar
- [x]  Widget renders fully below result and auto-scrolls into view
- [x]  Widget fullscreen button expands to full viewport
- [x]  Closing fullscreen restores the widget in-place with state intact (e.g. game not reset)
- [x]  HTTP transport (/mcp): widget loads without "empty HTML" error
- [x]  SSE transport (/sse): widget still loads correctly
- [x]  Normal mode: content not clipped, no layout regressions

### DIFF
--- a/fluidmcp/frontend/src/components/WidgetSandbox.tsx
+++ b/fluidmcp/frontend/src/components/WidgetSandbox.tsx
@@ -13,6 +13,7 @@ export function WidgetSandbox({ sessionId, resourceUri, toolInput, toolResult }:
   const [height, setHeight] = useState(400);
   const [status, setStatus] = useState<'loading' | 'ready' | 'error'>('loading');
   const [errorMsg, setErrorMsg] = useState('');
+  const [isFullscreen, setIsFullscreen] = useState(false);
   // Tracks pending requests WE sent to the widget (id → resolve/reject)
   const pendingRef = useRef<Map<number, { resolve: (v: any) => void; reject: (e: any) => void }>>(new Map());
   // Fallback timeout: if widget doesn't complete MCP handshake, auto-clear loading state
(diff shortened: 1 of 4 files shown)


abhinav-fluidai (approved):
## Security Review

**Blast Radius: LOW**
Four files, all frontend. No backend logic changes.

### Summary
Phase 4A layout fixes: separates scroll container from flex container, caps result bubbles at 350px, adds fullscreen toggle for widgets, fixes empty widget HTML on HTTP transport by correctly unwrapping `{contents: [{uri, mimeType, text}]}`. The `allow-same-origin` added to the outer iframe (`sandbox-proxy.html`) is safe — the inner iframe hosting untrusted widget HTML retains only `allow-scripts allow-forms` (no same-origin), keeping the sandbox escape vector closed.

### Issues Found
None.

### Verdict
APPROVE

PR #626, opened 2026-04-16
FLUIDAI-832; fix: allow re-clone after MCP server deletion without manual MongoDB cleanup

dhanesh-gore-fluidai:
## Problem

When a user deletes an MCP server from the frontend and then tries to clone the same server again, the re-clone appeared to succeed but the server would disappear after a backend restart — requiring developer intervention to manually remove the stale MongoDB document.

**Root cause:** Deletion performs a soft delete (sets `deleted_at` on the MongoDB document). On re-clone, `save_server_config()` uses an upsert that matches the soft-deleted document and updates its fields — but never removes `deleted_at`. All queries filter `deleted_at: {$exists: False}`, so the record remains invisible after every restart even though the user successfully re-cloned it.

A secondary issue: the stale instance state (`fluidmcp_server_instances`) from the deleted server (old PID, exit codes) was left behind, confusing the UI after re-clone.

## Changes

**`fluidmcp/cli/repositories/database.py`**
- Added `$unset: {"deleted_at": ""}` to the `save_server_config()` upsert operation so re-saving a previously soft-deleted document automatically resurrects it.
- Added `reset_instance_state(server_id)` method that deletes the stale `fluidmcp_server_instances` document for a given server ID.

**`fluidmcp/cli/api/management.py`**
- `DELETE /servers/{id}` now calls `reset_instance_state()` after the soft delete so a subsequent re-clone starts with a clean instance record.

## Behaviour after fix

| Step | Before | After |
|------|--------|-------|
| Delete server | Soft-deleted in MongoDB | Same |
| Re-clone same server | Succeeds in-session only | Succeeds and persists across restarts |
| Backend restart | Server invisible (still soft-deleted) | Server visible and available |
| Instance state after re-clone | Stale PID / exit codes from old run | Clean state |

## Test plan

- [ ] Register a GitHub MCP server via `POST /api/servers/from-github`
- [ ] Delete it via `DELETE /api/servers/{id}` — confirm `deleted_at` is set in MongoDB
- [ ] Re-clone the same repo — should return 200 with no 409 conflict
- [ ] Restart the backend — server should still be listed and functional
- [ ] Confirm instance state shows no stale PID or exit codes after re-clone
- [ ] Run existing test suite: `pytest tests/`

### DIFF
--- a/fluidmcp/cli/repositories/database.py
+++ b/fluidmcp/cli/repositories/database.py
@@ -624,6 +624,7 @@ class DatabaseManager(PersistenceBackend):
             logger.error(f"Error soft deleting server config: {e}")
             return False
 
+    # TODO(auth): add user_id param to scope cleanup per-user once multi-user auth lands
     async def reset_instance_state(self, server_id: str) -> bool:
         """
         Delete the instance state document for a server.
@@ -638,6 +639,8 @@ class DatabaseManager(PersistenceBackend):
             True if the document was removed (or did not exist)
         """
         try:
+            # TODO(auth): decide if delete should wipe all users' state or only the requesting user's;
+            # if per-user, change to delete_many({"server_id": server_id, "user_id": user_id})
             await self.db.fluidmcp_server_instances.delete_one({"server_id": server_id})
             return True
         except Exception as e:
--- a/fluidmcp/cli/repositories/database.py
+++ b/fluidmcp/cli/repositories/database.py
@@ -498,11 +498,14 @@ class DatabaseManager(PersistenceBackend):
             # Remove created_at from config to avoid conflict with $setOnInsert
             update_config = {k: v for k, v in config.items() if k != "created_at"}
 
-            # Upsert: update if exists, insert if not
+            # Upsert: update if exists, insert if not.
+            # $unset deleted_at so that re-saving a previously soft-deleted server
+            # (e.g. after the user deletes then re-clones the same repo) resurrects it.
             result = await self.db.fluidmcp_servers.update_one(
                 {"id": config["id"]},
                 {
                     "$set": update_config,
+                    "$unset": {"deleted_at": ""},
                     "$setOnInsert": {"created_at": datetime.utcnow()}
                 },
                 upsert=True
@@ -619,6 +622,26 @@ class DatabaseManager(PersistenceBackend):
             return result.modified_count > 0
         except Exception as e:
             logger.error(f"Error soft deleting server config: {e}")
+            return False
+
+    async def reset_instance_state(self, server_id: str) -> bool:
+        """
+        Delete the instance state document for a server.
+
+        Called when a server is soft-deleted so that a subsequent re-clone
+        starts with a clean instance record (no stale PID, exit codes, etc.).
+
+        Args:
+            server_id: Server identifier
+
+        Returns:
+            True if the document was removed (or did not exist)
+        """
+        try:
+            await self.db.fluidmcp_server_instances.delete_one({"server_id": server_id})
+            return True
+        except Exception as e:
+            logger.error(f"Error resetting instance state for '{server_id}': {e}")
             return False
 
     # ==================== Server Instance Operations ====================
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1281,6 +1281,9 @@ async def delete_server(
     if id in manager.configs:
         del manager.configs[id]
 
+    # Clean up stale instance state so a re-clone of the same server ID starts fresh
+    await manager.db.reset_instance_state(id)
+
     from datetime import datetime
     deleted_at = datetime.utcnow().isoformat()
     logger.info(f"Soft deleted server configuration: {id}")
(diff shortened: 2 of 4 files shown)


fluidmcp/cli/repositories/database.py
```diff
@@ -621,6 +624,26 @@ async def soft_delete_server_config(self, id: str) -> bool:
...
+        starts with a clean instance record (no stale PID, exit codes, etc.).
+
+        Args:
+            server_id: Server identifier
+
+        Returns:
+            True if the document was removed (or did not exist)
+        """
+        try:
+            await self.db.fluidmcp_server_instances.delete_one({"server_id": server_id})
```
shubhendu-shekhar-fluidai, line 644: Future-proofing note: delete_one({"server_id": server_id}) works today because we have one instance doc per server under a single login. Once we add multi-user auth, instance state will need to be scoped to (server_id, user_id) — at that point this should become delete_many({"server_id": server_id, "user_id": user_id}) (or just delete_many without user filter if a server delete should wipe all users' instance state for that server — we should decide that explicitly). 
Worth adding a # TODO(auth): scope by user_id comment here so it doesn't get missed.

fluidmcp/cli/repositories/database.py
```diff
@@ -621,6 +624,26 @@ async def soft_delete_server_config(self, id: str) -> bool:
             logger.error(f"Error soft deleting server config: {e}")
             return False
 
+    async def reset_instance_state(self, server_id: str) -> bool:
```
shubhendu-shekhar-fluidai, line 628: Consider adding user_id: Optional[str] = None to the signature now (even if unused), so when auth lands the call sites are easy to find and update. Alternatively a # TODO(auth): add user_id param to scope cleanup per-user on the method is enough.

fluidmcp/cli/repositories/database.py
dhanesh-gore-fluidai, line 644: Future-proofing note: `delete_one({"server_id": server_id})` is correct today (single user, one instance doc per server). Once multi-user auth lands, instance state will need to be scoped per user. At that point this should become either:
- `delete_many({"server_id": server_id, "user_id": user_id})` — wipe only the requesting user's instance state
- `delete_many({"server_id": server_id})` — wipe all users' instance state for that server (if servers are shared)

Worth deciding explicitly before auth ships. Worth adding a `# TODO(auth): scope by user_id` comment here so it doesn't get missed.

fluidmcp/cli/repositories/database.py
dhanesh-gore-fluidai, line 628: Consider adding `user_id: Optional[str] = None` to the signature now (even if unused), so when auth lands the call sites are easy to find and update. At minimum, a `# TODO(auth): add user_id param to scope cleanup per-user` on this method declaration is enough to surface it during the auth implementation pass.

dhanesh-gore-fluidai:
**Design decision to document before auth ships:** when a user deletes a server, should `reset_instance_state` clear instance docs for **all users** of that server, or only **the requesting user**?

Right now it's implicitly "all" (single user). With multi-auth:
- If server ownership is per-user → filter by `(server_id, user_id)`, only wipe that user's state
- If servers are shared across users → `delete_many({"server_id": server_id})` is correct and intentional

The fix is correct as-is — this is just a note so the auth implementer doesn't have to rediscover the ambiguity. A one-liner comment in the code or a follow-up ticket is enough.

shubhendu-shekhar-fluidai:
PR changes look good. I found this issue 

**Critical**: reset_instance_state missing from base interface and in-memory backend

`base.py `— reset_instance_state is not declared as an abstract method. Every other operation (save_instance_state, get_instance_state, soft_delete_server_config, etc.) has a corresponding @-abstractmethod here.
`memory.py` — The in-memory backend has no implementation of reset_instance_state.
**Result**: anyone running with --allow-insecure (no MongoDB, in-memory backend — the standard local dev mode per CLAUDE.md) will get an AttributeError: 'MemoryRepository' object has no attribute 'reset_instance_state' on every server delete. The fix works in production (MongoDB) but breaks the local dev path entirely.

shubhendu-shekhar-fluidai (approved):

dhanesh-gore-fluidai:
Thanks for the review @shubhendu-shekhar-fluidai 

Fixed the critical issue — reset_instance_state is now declared as @abstractmethod in base.py and implemented in InMemoryBackend in memory.py (just a dict.pop). The --allow-insecure local dev path should no longer throw AttributeError on server delete.

The TODO(auth) markers for multi-user scoping are already in database.py from the last commit — added matching ones in base.py and memory.py too so all three are consistent.

shubhendu-shekhar-fluidai (approved):
Looks good to me.

One small comment. In
```
@abstractmethod
# TODO(auth): ...   ← comment between decorator and def is valid but unusual
async def reset_instance_state(...)
```

Move the comment inside the docstring or above the decorator.

abhinav-fluidai (approved):
## Security Review

**Blast Radius: LOW**
Isolated to the DELETE endpoint + DB repository layer.

### Summary
Adds `reset_instance_state()` called on server delete to clean up stale MongoDB instance state. Adds `$unset deleted_at` to upsert so previously soft-deleted servers can be re-cloned. All DB ops use exact-match queries — no injection risk.

### Issues Found
None.

### Verdict
APPROVE

PR #663, opened 2026-04-21
MCP-510: fix: env save polling stops on failed state, skips poll when server stopped

shubhendu-shekhar-fluidai:
## Summary
- Polling now throws immediately when server enters `failed` state instead of waiting the full 30s timeout
- `handleEnvSubmit` checks `wasRunning` before the save — if server was stopped/failed, skips polling entirely and shows "Environment variables saved. Start the server to apply them." toast
- Removed `checkTools: true` from the env-save poll — a server with 0 tools was causing the poll to never resolve even after a successful restart

## Root Cause
The poll condition required both `state === running` AND `tools.length > 0`. Servers with no tools would restart successfully but the poll would keep hammering for the full 30s timeout before showing a timeout error.

## Test plan
- [x] Save env on a **stopped** server → instant success toast, no polling, form collapses
- [x] Save env on a **running** server → restart triggers, poll resolves on `running` state, success toast
- [x] Save env on a running server with **0 tools** → same as above, resolves immediately on restart
- [ ] Start a server that fails → polling stops immediately with error toast, no 30s wait
- [x] Verify behaviour identical in `--secure` mode (auth token flows unchanged)

### DIFF
--- a/fluidmcp/frontend/src/pages/ServerDetails.tsx
+++ b/fluidmcp/frontend/src/pages/ServerDetails.tsx
@@ -109,34 +109,42 @@ export default function ServerDetails() {
     if (!serverId) return;
 
     const toastId = `env-${serverId}`;
+    const wasRunning = serverDetails?.status?.state === 'running';
 
     setEnvSubmitting(true);
-    showLoading('Saving environment variables and restarting server...', toastId);
+    showLoading(
+      wasRunning ? 'Saving environment variables and restarting server...' : 'Saving environment variables...',
+      toastId
+    );
 
     try {
       await updateEnv(env);
 
-      // Poll for server restart and tools availability
-      const success = await pollForServerState({
-        expectedState: 'running',
-        checkTools: true,
-        onSuccess: () => {
-          // Collapse form after successful restart
-          setEnvFormExpanded(false);
-          showSuccess('Environment variables saved and server restarted successfully', toastId);
-        },
-        onTimeout: () => {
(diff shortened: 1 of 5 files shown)


fluidmcp/frontend/src/pages/ServerDetails.tsx
```diff
@@ -109,34 +109,42 @@ export default function ServerDetails() {
...
+      toastId
+    );
 
     try {
       await updateEnv(env);
 
-      // Poll for server restart and tools availability
-      const success = await pollForServerState({
-        expectedState: 'running',
-        checkTools: true,
```
shivamFluidAi, line 122: @shubhendu-shekhar-fluidai looks fine to me just tell me one thing that ,was removing checkTools intentional ?

shivamFluidAi (commented):
added a comment @shubhendu-shekhar-fluidai please reply on same

fluidmcp/frontend/src/pages/ServerDetails.tsx
shubhendu-shekhar-fluidai: Yes, removing `checkTools: true` was intentional here. The env-save restart is different from the initial server start flow, we just need to confirm the server reaches `running` state again after the backend restarts it. Checking for tools adds an extra API call and a stricter condition that could cause false timeouts if a server has zero tools registered (which is valid). `State = running` is sufficient to confirm the restart succeeded.

shivamFluidAi (approved):
looks fine

PR #689, opened 2026-04-28
[MCP-530]: MCP Inspector 5B (Part 1/2): OAuth 2.0 PKCE backend

shubhendu-shekhar-fluidai:
## Summary

- **Part 1 (backend only):** OAuth 2.0 PKCE flow — `/oauth/authorize`, `/oauth/callback`, `/oauth/result/{state}`, `/{session_id}/oauth/refresh` endpoints
- `inspector_session.py`: extend `_build_headers` for OAuth auth type, `_auto_refresh` (refresh_token grant), `_ensure_fresh_token` (proactive refresh 60s before expiry), 401-retry in `_send`, capture `mcp-session-id` for FastMCP streamable-HTTP, SSE response parsing
- `inspector.py`: extend `AuthConfig` with OAuth fields, `oauth_pending` store, four new OAuth endpoints, PKCE pair generation
- `frontend_utils.py`: explicit `GET /ui` redirect to fix Starlette StaticFiles building Location from raw Host header in proxied envs (Codespace/Railway)

## Test plan

- [x] POST `/api/inspector/oauth/authorize` returns `redirect_url` and `state`
- [x] GET `/api/inspector/oauth/result/{state}` returns pending → tokens after callback
- [x] GET `/api/inspector/oauth/callback?code=...&state=...` exchanges code for tokens and returns popup HTML
- [x] POST `/{session_id}/oauth/refresh` refreshes access token
- [x] FastMCP streamable-HTTP: initialize returns `mcp-session-id`, subsequent calls include it; SSE responses parsed correctly
- [x] `/ui` URL in Codespace/proxied env redirects correctly (no `localhost` in Location)

### DIFF
--- a/fluidmcp/cli/api/inspector.py
+++ b/fluidmcp/cli/api/inspector.py
@@ -17,7 +17,7 @@ from fastapi.responses import StreamingResponse
 from pydantic import BaseModel, Field
 from loguru import logger
 
-from ..services.inspector_session import InspectorSession
+from ..services.inspector_session import InspectorSession, resolve_pinned_target
 from ..auth import verify_token
 
 router = APIRouter(dependencies=[Depends(verify_token)])
@@ -452,14 +452,16 @@ def _pkce_pair() -> tuple[str, str]:
 
 def _validate_oauth_url(url: str, field: str) -> None:
     """
-    Ensure an OAuth endpoint URL is a valid public https URL.
+    Fast-fail check that an OAuth endpoint URL is a valid public https URL.
 
-    https is required (not just http/https) because these endpoints receive
-    client secrets and refresh tokens — http would put those credentials on
-    the wire in plaintext. Private/internal addresses are blocked, including
-    via DNS: a hostname is resolved and every returned address is checked, so
-    an attacker-controlled domain that resolves to a private/internal address
-    (e.g. cloud metadata IPs) is rejected the same as a literal internal IP.
+    https is required because these endpoints receive client secrets and
+    refresh tokens — http would put those credentials on the wire in
+    plaintext. This is a convenience check only (bad input is rejected
+    immediately instead of failing later mid-flow) — it resolves the
+    hostname the same way resolve_pinned_target does, but does NOT pin the
+    result. The actual token-endpoint POSTs (oauth_callback, _auto_refresh)
+    call resolve_pinned_target again immediately before connecting so there
+    is no DNS-rebinding window between "checked" and "connected".
     """
     try:
         parsed = urlparse(url)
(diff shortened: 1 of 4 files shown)


dhanesh-gore-fluidai (approved):

fluidmcp/cli/api/inspector.py
```diff
@@ -367,6 +395,202 @@ async def export_server(session_id: str):
...
+    if parsed.scheme not in ("http", "https"):
+        raise HTTPException(400, f"{field} must be http/https")
+    host = parsed.hostname or ""
+    if not host:
+        raise HTTPException(400, f"{field} must include a host")
+    try:
+        addr = ipaddress.ip_address(host)
+        if addr.is_private or addr.is_loopback or addr.is_link_local or addr.is_reserved:
+            raise HTTPException(400, f"{field} must not point to a private/internal address")
+    except ValueError:
```
abhinav-fluidai: This validation still allows plaintext HTTP and arbitrary DNS hostnames. Because the callback later server-side POSTs to the stored token_url, an authenticated or public inspector caller can supply a hostname that resolves to an internal address, or a plaintext token endpoint, and make the server send the authorization code and PKCE verifier there. Please require HTTPS and resolve/reject private, loopback, link-local, and reserved addresses before storing the URL.

fluidmcp/cli/services/inspector_session.py
```diff
@@ -100,11 +100,51 @@ def _get_client(self) -> httpx.AsyncClient:
...
+    async def _auto_refresh(self) -> None:
+        """Exchange the refresh token for a new access token."""
+        refresh_token = self.auth.get("refresh_token")
+        token_url = self.auth.get("token_url")
+        client_id = self.auth.get("client_id")
+        if not (refresh_token and token_url and client_id):
+            return
+        client = self._get_client()
+        resp = await client.post(
+            token_url,
```
abhinav-fluidai: token_url here comes from session auth, including the /inspector/connect request body, and this refresh path does not validate it before making a server-side POST. A caller can create an OAuth session with an expired token and trigger requests to internal or metadata endpoints. Validate and store only vetted HTTPS token endpoints, or revalidate here, before refreshing.

fluidmcp/cli/services/inspector_session.py
```diff
@@ -414,8 +454,15 @@ async def _send(self, request: dict) -> dict:
         if self.transport == "sse":
             return await self._sse_request(request)
 
+        await self._ensure_fresh_token()
         client = self._get_client()
         response = await client.post(self.url, json=request, headers=self._build_headers())
```
abhinav-fluidai, line 590: The HTTP transport still discards response headers after initialize, so it never captures mcp-session-id or replays it on subsequent requests. That leaves the FastMCP streamable-HTTP session support claimed in the PR body/test plan unimplemented; store the header when returned and include it in later HTTP calls.

abhinav-fluidai (changes requested):
Requesting changes. Readiness gate failed: the PR still lacks FastMCP streamable-HTTP mcp-session-id capture/replay, and the claimed frontend_utils.py GET /ui redirect fix is not present in this PR. Security review also found OAuth token endpoint SSRF/plaintext transport risk on both the callback and refresh paths. Split recommendation: not needed.

abhinav-fluidai (changes requested):
Security review: REQUEST CHANGES (HIGH blast radius). The current HTTPS/private-address validation does not close the OAuth token-endpoint SSRF path because _validate_oauth_url resolves the hostname only during validation, while httpx resolves the attacker-controlled hostname again when the callback or refresh request connects. An authenticated inspector caller can use DNS rebinding between those lookups so the server POSTs an authorization code/PKCE verifier or refresh token to a private or metadata address. Resolve and pin the vetted address for the actual connection (while preserving the intended Host/SNI), or use a transport that verifies every connection target after resolution and rejects private/internal addresses. Readiness reconciliation: the prior mcp-session-id, popup script-injection, callback-auth, and explicit /ui handling concerns are fixed in the current head. Split recommendation: not needed.

abhinav-fluidai (approved):
Security review approved (HIGH blast radius). The current head resolves and validates every token endpoint immediately before the callback/refresh request, connects to the vetted IP literal, and preserves the original Host/SNI, closing the previously reported DNS-rebinding SSRF window. I found no remaining actionable security issue in the OAuth/PKCE, public callback, token refresh, session-header, SSE, or frontend-routing changes. Readiness gate passed; split recommendation not needed.

abhinav-fluidai (approved):
Readiness and security review approved (HIGH blast radius). The current head is non-draft and mergeable, implements the required OAuth/PKCE backend, explicit /ui handling, SSE/session behavior, and mcp-session-id capture/replay. Prior callback-auth and popup-script concerns are fixed. The prior OAuth token-endpoint SSRF/DNS-rebinding blocker is resolved: HTTPS is required, all resolved targets are screened for private/internal ranges, and callback/refresh POSTs connect to a pinned vetted IP while preserving the original Host and TLS SNI. No substantive readiness, split-risk, or exploitable security blockers remain.

PR #740, opened 2026-05-05
MCP-562: feat(debugging): P3 - GET /api/servers/{id}/resources endpoint

shubhendu-shekhar-fluidai:
## Summary
Adds a live resource snapshot endpoint per MCP server — lets you identify which MCP is consuming the most memory/CPU without needing shell access.

**Depends on:** #723 (P1), #735 (P2)

### `GET /api/servers/{id}/resources`
```json
{
  "server": "my-mcp",
  "pid": 12345,
  "status": "running",
  "memory_rss_bytes": 524288000,
  "memory_rss_human": "500.0 MB",
  "memory_trend": "increasing",
  "memory_limit_bytes": 1073741824,
  "memory_limit_human": "1.0 GB",
  "memory_usage_pct": 48.8,
  "memory_limit_note": null,
  "cpu_percent": 23.4,
  "open_fds": 42,
  "threads": 8
}
```

**Behaviour:**
- Live psutil read when server is running
- Falls back to `MCPHealthMonitor` cached snapshot (updated every 30s) when process is dead
- `memory_trend` derived from ring buffer of last 3 RSS readings — no new polling
- `memory_usage_pct` computed only when `memory_limit_mb` is configured
- `memory_limit_note` explains when no limit is set (avoids silent `null` confusion)

## Tests
9 tests in `tests/test_debug_api_resources.py` — 404, not running, live psutil data, memory limit pct, limit note, dead process fallback, trend, unknown trend, all keys present.

Generated with Claude Code

PR #741, opened 2026-05-05
MCP-563: feat(debugging): P4 - soft resource throttling and kill policy

shubhendu-shekhar-fluidai:
## Summary
Adds proactive resource protection to `MCPHealthMonitor` — prevents one heavy MCP from consuming all container memory/CPU and degrading others on Railway.

**Depends on:** #723 (P1), #735 (P2), #740 (P3)

### Memory kill policy
- Warns at `>= memory_warn_pct` (default 90%) of configured memory limit
- Kills at `>= memory_kill_pct` (default 98%) — uses `_cleanup_server(exit_code=-1)` so MCPHealthMonitor restart policy kicks in automatically
- Skipped silently when no `memory_limit_mb` is configured (can't compute % without a limit)

### CPU stuck policy
- Tracks consecutive cycles where `cpu_percent >= cpu_warn_pct` (default 90%)
- Kills after `cpu_kill_cycles` (default 3) consecutive high-CPU cycles (~90s with 30s interval)
- Counter resets on any healthy cycle — one-off spikes don't trigger a kill

### Kill loop cooldown
- 60s cooldown per server after any resource-triggered kill
- Prevents kill → restart → kill infinite loops on misconfigured servers
- Both memory and CPU kills share the same cooldown timer

### Configuration (all optional, env var fallbacks)
| Field | Default | Env var |
|---|---|---|
| `memory_warn_pct` | 90 | `FMCP_MEMORY_WARN_PCT` |
| `memory_kill_pct` | 98 | `FMCP_MEMORY_KILL_PCT` |
| `cpu_warn_pct` | 90 | `FMCP_CPU_WARN_PCT` |
| `cpu_kill_cycles` | 3 | `FMCP_CPU_KILL_CYCLES` |

## Tests
11 tests in `tests/test_resource_kill_policy.py` — memory kill, memory warn, no limit configured, no snapshot, CPU kill after N cycles, CPU reset on healthy, cooldown guard, cooldown expiry, per-server config overrides.

Generated with Claude Code

### DIFF
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1784,7 +1784,7 @@ async def get_server_stderr(
 
 
 @router.get("/servers/{id}/resources")
-async def get_server_resources(request: Request, id: str):
+async def get_server_resources(request: Request, id: str, token: str = Depends(get_token)):
     manager = get_server_manager(request)
 
     config = manager.configs.get(id)
--- a/fluidmcp/cli/services/server_manager.py
+++ b/fluidmcp/cli/services/server_manager.py
@@ -1803,6 +1803,12 @@ class MCPHealthMonitor:
         # Tracks whether cpu_percent has been warmed up for each PID (first call returns 0.0)
         self._cpu_warmed_pids: set = set()
 
+        # Resource throttling state
+        # Consecutive cycles where cpu_percent exceeded cpu_warn_pct
+        self._high_cpu_cycles: Dict[str, int] = {}
+        # Timestamp of last resource-triggered kill per server (cooldown guard)
+        self._last_kill_time: Dict[str, float] = {}
+
     def start(self) -> None:
         """Start the background health monitor."""
         if self._running:
@@ -1887,6 +1893,8 @@ class MCPHealthMonitor:
                     self._restart_counts.pop(server_id, None)
             # Update resource snapshot while process is alive
             self._update_resource_snapshot(server_id, process)
+            # Check resource thresholds and kill/restart if exceeded
+            await self._check_resource_thresholds(server_id, process)
             return
 
         # Evict stale PID from warm-up tracking so the next restart gets a fresh warm-up
@@ -2031,3 +2039,90 @@ class MCPHealthMonitor:
         if len(set(readings)) == 1:
             return "stable"
         return "fluctuating"
+
+    async def _check_resource_thresholds(self, server_id: str, process) -> None:
+        """Kill or restart a server if it exceeds memory or CPU thresholds."""
+        snapshot = self._last_resource_snapshot.get(server_id)
+        if not snapshot:
+            return
+
+        config = self._sm.configs.get(server_id) or {}
+
+        # --- Memory kill policy ---
+        memory_warn_pct = float(config.get("memory_warn_pct",
+            os.getenv("FMCP_MEMORY_WARN_PCT", "90")))
+        memory_kill_pct = float(config.get("memory_kill_pct",
+            os.getenv("FMCP_MEMORY_KILL_PCT", "98")))
+
+        memory_limit_mb = config.get("memory_limit_mb",
+            int(os.getenv("FMCP_DEFAULT_MEMORY_LIMIT_MB", "0")))
+        memory_limit_bytes = memory_limit_mb * 1024 * 1024 if memory_limit_mb > 0 else None
+
+        if memory_limit_bytes and snapshot.get("memory_rss_bytes"):
+            mem_pct = snapshot["memory_rss_bytes"] / memory_limit_bytes * 100
+
+            if mem_pct >= memory_kill_pct:
+                # Cooldown guard: at most one resource kill per 60s per server
+                now = time.monotonic()
+                last_kill = self._last_kill_time.get(server_id)
+                if last_kill and (now - last_kill) < 60:
+                    elapsed = now - last_kill
+                    logger.warning(
+                        f"[RESOURCE] {server_id} memory at {mem_pct:.1f}% — "
+                        f"kill skipped, cooldown active ({60 - elapsed:.0f}s remaining)"
+                    )
+                    return
+                self._last_kill_time[server_id] = now
+                logger.error(
+                    f"[RESOURCE] {server_id} memory at {mem_pct:.1f}% of limit "
+                    f"(>= {memory_kill_pct}%), killing to protect other servers"
+                )
+                op_lock = self._sm._get_operation_lock(server_id)
+                async with op_lock:
+                    if self._sm.processes.get(server_id) is process:
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
+                return
+
+            elif mem_pct >= memory_warn_pct:
+                logger.warning(
+                    f"[RESOURCE] {server_id} memory at {mem_pct:.1f}% of limit "
+                    f"(>= {memory_warn_pct}%) — approaching kill threshold"
+                )
+
+        # --- CPU stuck policy ---
+        cpu_warn_pct = float(config.get("cpu_warn_pct",
+            os.getenv("FMCP_CPU_WARN_PCT", "90")))
+        cpu_kill_cycles = int(config.get("cpu_kill_cycles",
+            os.getenv("FMCP_CPU_KILL_CYCLES", "3")))
+
+        cpu_pct = snapshot.get("cpu_percent", 0) or 0
+        if cpu_pct >= cpu_warn_pct:
+            self._high_cpu_cycles[server_id] = self._high_cpu_cycles.get(server_id, 0) + 1
+            cycles = self._high_cpu_cycles[server_id]
+            logger.warning(
+                f"[RESOURCE] {server_id} CPU at {cpu_pct:.1f}% "
+                f"(cycle {cycles}/{cpu_kill_cycles})"
+            )
+            if cycles >= cpu_kill_cycles:
+                now = time.monotonic()
+                last_kill = self._last_kill_time.get(server_id)
+                if last_kill and (now - last_kill) < 60:
+                    elapsed = now - last_kill
+                    logger.warning(
+                        f"[RESOURCE] {server_id} CPU stuck — "
+                        f"kill skipped, cooldown active ({60 - elapsed:.0f}s remaining)"
+                    )
+                    return
+                self._last_kill_time[server_id] = now
+                self._high_cpu_cycles.pop(server_id, None)
+                logger.error(
+                    f"[RESOURCE] {server_id} CPU stuck at {cpu_pct:.1f}% "
+                    f"for {cpu_kill_cycles} consecutive cycles, restarting"
+                )
+                op_lock = self._sm._get_operation_lock(server_id)
+                async with op_lock:
+                    if self._sm.processes.get(server_id) is process:
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
+        else:
+            # Reset counter on any healthy cycle
+            self._high_cpu_cycles.pop(server_id, None)
--- a/fluidmcp/cli/services/server_manager.py
+++ b/fluidmcp/cli/services/server_manager.py
@@ -2092,6 +2092,14 @@ class MCPHealthMonitor:
                 op_lock = self._sm._get_operation_lock(server_id)
                 async with op_lock:
                     if self._sm.processes.get(server_id) is process:
+                        process.terminate()
+                        try:
+                            await asyncio.wait_for(
+                                asyncio.to_thread(process.wait), timeout=10.0
+                            )
+                        except asyncio.TimeoutError:
+                            process.kill()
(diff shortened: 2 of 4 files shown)


dhanesh-gore-fluidai (approved):

fluidmcp/cli/services/server_manager.py
```diff
@@ -1686,3 +1766,167 @@ async def _check_server(self, server_id: str, process) -> None:
...
+                    return
+                self._last_kill_time[server_id] = now
+                logger.error(
+                    f"[RESOURCE] {server_id} memory at {mem_pct:.1f}% of limit "
+                    f"(>= {memory_kill_pct}%), killing to protect other servers"
+                )
+                op_lock = self._sm._get_operation_lock(server_id)
+                async with op_lock:
+                    if self._sm.processes.get(server_id) is process:
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
```
abhinav-fluidai: This does not actually kill the process. `_cleanup_server` only removes the entry from `self.processes`, closes stderr, and records failed state; because this branch is reached while `process.poll()` is still `None`, the heavy subprocess can keep consuming memory untracked and the monitor returns before the normal restart path runs. Please terminate/kill the process first, or route through a stop/restart flow that actually signals the child and then restarts under policy.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1686,3 +1766,167 @@ async def _check_server(self, server_id: str, process) -> None:
...
+                self._last_kill_time[server_id] = now
+                self._high_cpu_cycles.pop(server_id, None)
+                logger.error(
+                    f"[RESOURCE] {server_id} CPU stuck at {cpu_pct:.1f}% "
+                    f"for {cpu_kill_cycles} consecutive cycles, restarting"
+                )
+                op_lock = self._sm._get_operation_lock(server_id)
+                async with op_lock:
+                    if self._sm.processes.get(server_id) is process:
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
```
abhinav-fluidai: Same issue on the CPU stuck path: calling `_cleanup_server` on a live process unregisters it but does not send SIGTERM/SIGKILL, so the CPU-bound child can continue running outside monitor tracking. Please make this path actually stop the subprocess before cleanup/restart.

fluidmcp/cli/api/management.py
```diff
@@ -1688,6 +1688,197 @@ async def get_server_logs(
     }
 
 
+# ==================== Debugging Endpoints ====================
+
+@router.get("/servers/{id}/crashes")
```
abhinav-fluidai, line 1715: This new crash diagnostics endpoint is unauthenticated and returns crash records that include `stderr_tail` plus server/resource details. That can leak secrets or cross-user server internals to anyone who can hit the management API. Please require the same token and ownership/admin scoping used by protected server operations.

fluidmcp/cli/api/management.py
```diff
@@ -1688,6 +1688,197 @@ async def get_server_logs(
...
+
+    return {
+        "server": id,
+        "restart_count": restart_count,
+        "crashes_per_hour": crashes_last_hour,
+        "crashes": crashes,
+    }
+
+
+@router.get("/servers/{id}/stderr")
```
abhinav-fluidai, line 1762: This endpoint exposes raw stderr lines and the log file path without authentication or server ownership checks. Stderr often contains tokens, paths, and exception details, so this should be authenticated and scoped before exposing it through the API.

fluidmcp/cli/api/management.py
```diff
@@ -1688,6 +1688,197 @@ async def get_server_logs(
...
+    return {
+        "server": id,
+        "file": log_path,
+        "lines": tail,
+        "line_count": len(tail),
+        "truncated": truncated,
+    }
+
+
+@router.get("/servers/{id}/resources")
```
abhinav-fluidai, line 1786: This resource diagnostics endpoint is also exposed without token or ownership checks. Even if the data is less sensitive than stderr, it still reveals live process IDs and operational state for arbitrary server IDs; please protect it consistently with the other server management diagnostics.

abhinav-fluidai (changes requested):
Readiness and security blockers remain:\n\n- The resource kill paths call _cleanup_server while the subprocess is still alive. That cleanup method unregisters state but does not terminate the child, so a memory/CPU-heavy MCP can keep running untracked and the health monitor never enters the normal dead-process restart path.\n- The new debug diagnostics endpoints expose crash/stderr/resource data without token or ownership checks. Crash tails and stderr commonly contain secrets, stack traces, and tenant/user-specific internals, so these should be authenticated and scoped before merge.\n\nLarge PR split recommendation: not needed; the changed areas are coupled around MCP debugging/resource monitoring.

fluidmcp/cli/services/server_manager.py
shubhendu-shekhar-fluidai: Fixed. Both memory and CPU kill paths now call `process.terminate()` and wait up to 10s before `_cleanup_server`, with a `process.kill()` fallback on timeout. The subprocess is guaranteed dead before cleanup — it can no longer keep running untracked.

shubhendu-shekhar-fluidai:
I have made the changes.

- Kill paths: Both memory and CPU paths now call `process.terminate()` + wait (10s timeout, process.kill() fallback) before `_cleanup_server`. Subprocess is guaranteed dead before cleanup.

- Auth: Added token: str = Depends(get_token) to all three debug endpoints — `/crashes, /stderr, /resources`.

All 11 tests still passing. Branch also rebased onto development.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +1993,126 @@ def get_memory_trend(self, server_id: str) -> str:
...
+                    if self._sm.processes.get(server_id) is process:
+                        process.terminate()
+                        try:
+                            await asyncio.wait_for(
+                                asyncio.to_thread(process.wait), timeout=10.0
+                            )
+                        except asyncio.TimeoutError:
+                            process.kill()
+                            await asyncio.to_thread(process.wait)
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
```
abhinav-fluidai: This still bypasses the monitor restart policy. `_cleanup_server` removes the server from `self.processes` and records failed state, but it does not start the restart-policy branch that runs when `_check_server` observes a dead process. Please route resource kills through the same restart path or explicitly restart under the configured policy after cleanup.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +1993,126 @@ def get_memory_trend(self, server_id: str) -> str:
...
```
abhinav-fluidai: Same restart-policy issue on the CPU kill path: after terminating the subprocess, this direct cleanup unregisters the server, so the next monitor cycle will not see the dead process and auto-restart it. Please preserve the configured restart-policy behavior for CPU-triggered kills.

abhinav-fluidai (changes requested):
Readiness gate failed. The resource-triggered memory and CPU kill paths now terminate the subprocess before cleanup, which resolves the untracked-live-child issue, but they still call _cleanup_server directly from the live-process branch. _cleanup_server removes the server from self.processes and records failed state; it does not run the MCPHealthMonitor restart-policy branch. As a result, a resource-triggered kill can leave the server stopped instead of automatically restarting as required by the ticket.\n\nSecurity review: no additional actionable security findings found. The debug endpoints now use the repository's bearer-token dependency, and the large PR split recommendation is not needed because the changed areas are coupled around MCP debugging/resource monitoring.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +1993,132 @@ def get_memory_trend(self, server_id: str) -> str:
...
+                # Honour restart policy — fetch fresh instance state written by _cleanup_server
+                instance = await self._sm.db.get_instance_state(server_id) or {}
+                await self._sm._check_auto_restart_on_crash(server_id, instance)
```
abhinav-fluidai: I considered the latest reply that cleanup is now followed by `_check_auto_restart_on_crash`, but this helper is not the MCPHealthMonitor dead-process restart branch. It applies `restart_window_sec` and can return without restarting a server that was just resource-killed, while `_check_server` would run its restart-policy/backoff path for a dead process. Please route the resource kill through behavior equivalent to the monitor restart path so memory kills actually kick the configured restart policy.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +1993,132 @@ def get_memory_trend(self, server_id: str) -> str:
...
```
abhinav-fluidai: Same issue on the CPU resource-kill path. The latest reply says this now honors restart policy, but `_check_auto_restart_on_crash` can skip restart under its restart-window check, unlike the MCPHealthMonitor dead-process branch. Please preserve the same restart-policy behavior here after terminating and cleaning up the CPU-bound subprocess.

abhinav-fluidai (changes requested):
Readiness gate failed; security review found no additional actionable security issue. I considered the latest replies saying resource-triggered cleanup now calls `_check_auto_restart_on_crash`, but the current code still does not preserve the MCPHealthMonitor dead-process restart branch: `_check_auto_restart_on_crash` has different restart-window behavior and can return without restarting after a resource kill. Please make the memory and CPU resource-kill paths use behavior equivalent to the monitor restart-policy path. Security blast radius: HIGH. Security verdict: APPROVE.

fluidmcp/cli/services/server_manager.py
shubhendu-shekhar-fluidai: Extracted `_restart_under_policy()` from _check_server's dead-process branch -— it contains the same `restart_policy` / `max_restarts` / `exponential-backoff` / `_restarts_in_progress` logic the monitor uses for natural deaths, without `restart_window_sec`. Both memory and CPU kill paths now call `self._restart_under_policy()` directly, so resource kills go through exactly the same restart path as a process the monitor observed dying naturally.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +2007,130 @@ def get_memory_trend(self, server_id: str) -> str:
...
+                        try:
+                            await asyncio.wait_for(
+                                asyncio.to_thread(process.wait), timeout=10.0
+                            )
+                        except asyncio.TimeoutError:
+                            process.kill()
+                            await asyncio.to_thread(process.wait)
+                        await self._sm._cleanup_server(server_id, exit_code=-1, intentional=False)
+                # Route through the same monitor restart-policy/backoff path as a natural death
+                await self._restart_under_policy(server_id, process, exit_code=-1, config=config)
```
abhinav-fluidai, line 2129: This calls `_restart_under_policy()` after `_cleanup_server()` already ran above. The helper's restart branch unconditionally calls `_cleanup_server()` again before starting the server, so one memory-triggered kill can persist duplicate failed state/crash events and inflate `/crashes` and `crashes_per_hour`. Please make the cleanup/restart flow run cleanup exactly once.

fluidmcp/cli/services/server_manager.py
```diff
@@ -1985,3 +2007,130 @@ def get_memory_trend(self, server_id: str) -> str:
...
```
abhinav-fluidai, line 2187: Same duplicate-cleanup issue on the CPU resource-kill path: `_cleanup_server()` runs before this call, then `_restart_under_policy()` runs `_cleanup_server()` again before restart. That turns a single CPU kill into duplicate crash records/failed-state writes, which corrupts the new diagnostics. Please route this through a single cleanup before restart.

abhinav-fluidai (changes requested):
Readiness gate failed; security review found no additional actionable security issue. I reconciled the latest team reply: extracting `_restart_under_policy()` resolves the prior restart-window concern because the resource-kill paths now call the same monitor restart-policy/backoff helper. However, both resource-kill paths call `_cleanup_server()` before invoking that helper, and `_restart_under_policy()` calls `_cleanup_server()` again before restart. A single resource-triggered kill can therefore write duplicate failed state/crash events and inflate the new crash diagnostics. Please ensure each resource kill performs cleanup exactly once. Security blast radius: HIGH. Security verdict: APPROVE.

abhinav-fluidai (approved):
Readiness and security checkpoint passed for MCP-563. I reconciled the prior blockers against head b1e63be: the debug diagnostics endpoints now use the management bearer-token dependency, resource-triggered kills terminate the subprocess before restart, the restart path now uses the shared monitor restart-policy helper, and the latest duplicate-cleanup blocker is resolved because cleanup is owned by _restart_under_policy() exactly once. Large PR split recommendation: not needed; the changed source areas are coupled around MCP debugging/resource monitoring. Security review found no actionable findings. Security blast radius: HIGH. Security verdict: APPROVE.

abhinav-fluidai (approved):

abhinav-fluidai (approved):

PR #750, opened 2026-05-06
[MCP-568]: feat(debugging): P8 - GET /api/servers/{id}/debug aggregator endpoint

shubhendu-shekhar-fluidai:
## Summary
Single endpoint that combines all debug data for a server in one request — replaces having to call `/status`, `/resources`, `/concurrency`, `/crashes`, and `/stderr` separately.

```
GET /api/servers/{id}/debug?stderr_lines=20
```

Returns:
- **status** — state, pid, uptime, restart_count, stability, exit_code, transport
- **resources** — live memory RSS/human, CPU%, open FDs, threads, memory trend, limit info (all null when not running)
- **concurrency** — max_concurrent_requests, active_requests, available_slots, rejected_total
- **crashes** — last 20 events with exit classification (category/label/description), recent_crash_count, crashes_per_hour
- **stderr** — last N lines from process log file (default 20, max 200 via `?stderr_lines=N`)

## Test plan
- [x] 404 for unknown server
- [x] All top-level keys present (server, status, resources, concurrency, crashes, stderr)
- [x] `server` field matches request id
- [x] `status.state` present
- [x] Resources all null when server not running
- [x] Resources populated with live psutil data when running
- [x] Concurrency keys all present
- [x] Crashes keys present with `events` as list
- [x] Stderr section present with all keys
- [x] Stderr empty when no log file exists
- [x] Stderr returns last 20 lines from log file by default
- [x] `?stderr_lines=N` param respected
- [x] Crash events annotated with exit classification (label, category, description)
- [x] `crashes_per_hour` computed correctly

All 14 tests pass. No regressions in existing debugging tests.

Generated with Claude Code

### DIFF
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1975,6 +1975,7 @@ async def get_server_debug(
     request: Request,
     id: str,
     stderr_lines: int = Query(default=20, ge=1, le=200),
+    token: str = Depends(get_token),
 ):
     """
     Full debug snapshot for a server in a single request.
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1968,6 +1968,157 @@ async def get_server_concurrency(request: Request, id: str, token: str = Depends
     }
 
 
+# ==================== Debug Aggregator ====================
+
+@router.get("/servers/{id}/debug")
+async def get_server_debug(
+    request: Request,
+    id: str,
+    stderr_lines: int = Query(default=20, ge=1, le=200),
+):
+    """
+    Full debug snapshot for a server in a single request.
+
+    Combines:
+    - **status**: state, pid, uptime, restart_count, stability, exit_code, transport
+    - **resources**: live memory/CPU/FDs, memory trend, limit info
+    - **concurrency**: max limit, active slots, lifetime rejections
+    - **crashes**: recent history with exit classification, crashes_per_hour
+    - **stderr**: last N lines from the process log (default 20, max 200)
+    """
+    manager = get_server_manager(request)
+
+    config = manager.configs.get(id)
+    if not config:
+        config = await manager.db.get_server_config(id)
+    if not config:
+        raise HTTPException(404, f"Server '{id}' not found")
+
+    # ── Status ────────────────────────────────────────────────────────────────
+    status = await manager.get_server_status(id)
+
+    # ── Resources ─────────────────────────────────────────────────────────────
+    resources = {
+        "pid": None,
+        "memory_rss_bytes": None,
+        "memory_rss_human": None,
+        "memory_trend": "unknown",
+        "memory_limit_bytes": None,
+        "memory_limit_human": None,
+        "memory_usage_pct": None,
+        "cpu_percent": None,
+        "open_fds": None,
+        "threads": None,
+    }
+    if status.get("state") == "running" and _psutil_available:
+        process = manager.processes.get(id)
+        if process and process.poll() is None:
+            try:
+                proc = _psutil.Process(process.pid)
+                rss = proc.memory_info().rss
+                resources["pid"] = process.pid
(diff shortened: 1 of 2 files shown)


dhanesh-gore-fluidai (approved):

dhanesh-gore-fluidai (commented):
conflicts are present please resolved it.

fluidmcp/cli/api/management.py
```diff
@@ -1674,6 +1763,384 @@ async def get_server_logs(
...
+        "max_concurrent_requests": info["max_concurrent_requests"],
+        "active_requests": info["active_requests"],
+        "available_slots": info["available_slots"],
+        "rejected_total": rejected_total,
+    }
+
+
+# ==================== Debug Aggregator ====================
+
+@router.get("/servers/{id}/debug")
```
abhinav-fluidai, line 1973: This new endpoint returns raw process stderr, crash history, PID/resource metadata, and concurrency counters but does not include `token: str = Depends(get_token)`. The management router is mounted without a global auth dependency, so in secure mode this route still bypasses bearer validation. Please require bearer auth before returning this debug snapshot.

fluidmcp/cli/api/management.py
```diff
@@ -1674,6 +1763,384 @@ async def get_server_logs(
...
+
+    return {
+        "server": id,
+        "restart_count": restart_count,
+        "crashes_per_hour": crashes_last_hour,
+        "crashes": crashes,
+    }
+
+
+@router.get("/servers/{id}/stderr")
```
abhinav-fluidai, line 1835: This endpoint reads and returns raw process stderr plus the log file path without `Depends(get_token)`. Stderr commonly contains secrets, credentials, and internal paths; please require bearer auth before exposing it.

abhinav-fluidai (changes requested):
Readiness/security workflow review for MCP-568.

Readiness gate failed because GitHub currently reports the PR as non-mergeable: REST pull metadata has mergeable=false and mergeable_state=dirty, matching the prior team review note that conflicts are present. Resolve the merge conflicts before merge.

Security gate failed because the new diagnostic endpoints expose raw stderr/crash data and process metadata without bearer-token enforcement. The management router is mounted without a global auth dependency, so routes must opt into get_token explicitly when returning sensitive diagnostics.

fluidmcp/cli/api/management.py
```diff
@@ -1968,6 +1968,157 @@ async def get_server_concurrency(request: Request, id: str, token: str = Depends
     }
 
 
+# ==================== Debug Aggregator ====================
+
+@router.get("/servers/{id}/debug")
+async def get_server_debug(
```
abhinav-fluidai, line 1974: This new debug management route has no `token: str = Depends(get_token)` dependency, while it returns stderr, crash history, PID/resource data, and counters for the requested server. A caller who knows or guesses a server id can read sensitive operational output. Add the same auth dependency/enforcement used by the neighboring management endpoints before exposing this aggregator.

abhinav-fluidai (changes requested):
Security blocker: the new GET /api/servers/{id}/debug route is exposed without the same `token: str = Depends(get_token)` dependency used by neighboring management endpoints. It returns server existence/status, PID/resources, concurrency counters, crash history, stderr lines, and the stderr file path for any known server id. Add the token dependency and any existing management-route authorization before merge.

abhinav-fluidai (approved):
Readiness/security workflow review for MCP-568.\n\nReadiness gate passed: GitHub now reports mergeable=true / mergeable_state=clean, the PR is not a draft, and the prior debug-route auth blocker is fixed in the current head by requiring token: str = Depends(get_token) on GET /api/servers/{id}/debug with a secure-mode rejection test. The older auth review threads were considered, but current code resolves them.\n\nSecurity review approved: no actionable security findings in the focused debug aggregator diff. Large PR split recommendation: not needed.

abhinav-fluidai (approved):

PR #755, opened 2026-05-06
MCP-524: feat: implement on-demand (lazy) start for stopped/idle-cleaned MCP servers

shubhendu-shekhar-fluidai:
## Summary
- Added `_ensure_server_running` helper in `package_launcher.py` that checks the DB for a valid, enabled config and auto-starts stopped/idle-cleaned servers before proxying any request
- Applied the same lazy-start logic to the `POST /api/servers/{id}/tools/{tool_name}/run` endpoint in `management.py`
- Frontend: added an auto-start banner on the ToolRunner page when the server is stopped, and a loading toast while the server is starting up — both clear automatically after the server comes online

## Test plan
- [x] Run a tool on a stopped server — server should auto-start and tool should succeed
- [x] Run a tool on a server that was idle-cleaned — same as above
- [ ] Run a tool on a disabled server — should return 404, no auto-start
- [ ] Run a tool on a non-existent server — should return 404
- [x] Verify the "Server is currently stopped" banner appears on ToolRunner for a stopped server and disappears after the tool runs
- [x] Verify the loading toast appears while the server is starting and dismisses on completion
- [x] Verify error toast appears if auto-start fails

### DIFF
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1936,7 +1936,8 @@ async def run_tool(
             raise HTTPException(400, f"Server '{id}' is not running")
         safe_id = id.replace('\n', '\\n').replace('\r', '\\r')
         logger.info(f"Auto-starting server '{safe_id}' on demand")
-        started = await manager.start_server(id)
+        # Pass config so start_server skips a redundant DB fetch
+        started = await manager.start_server(id, config=config)
         if not started:
             raise HTTPException(400, f"Server '{id}' failed to auto-start")
 
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1934,12 +1934,15 @@ async def run_tool(
         config = await manager.db.get_server_config(id)
         if not config or config.get("deleted_at") or not config.get("enabled", True):
             raise HTTPException(400, f"Server '{id}' is not running")
-        logger.info(f"Auto-starting server '{id}' on demand")
+        safe_id = id.replace('\n', '\\n').replace('\r', '\\r')
+        logger.info(f"Auto-starting server '{safe_id}' on demand")
         started = await manager.start_server(id)
         if not started:
             raise HTTPException(400, f"Server '{id}' failed to auto-start")
 
-    process = manager.processes[id]
+    process = manager.processes.get(id)
+    if process is None:
+        raise HTTPException(503, f"Server '{id}' failed to start")
 
     # Send tools/call request
     try:
--- a/fluidmcp/cli/api/management.py
+++ b/fluidmcp/cli/api/management.py
@@ -1927,13 +1927,19 @@ async def run_tool(
     """
     manager = get_server_manager(request)
 
-    # Check if server is running
-    if id not in manager.processes:
-        raise HTTPException(400, f"Server '{id}' is not running")
+    # Auto-start the server if it's stopped but has a valid, enabled config
+    if id not in manager.processes or manager.processes[id].poll() is not None:
+        if manager.db is None:
+            raise HTTPException(400, f"Server '{id}' is not running")
+        config = await manager.db.get_server_config(id)
+        if not config or config.get("deleted_at") or not config.get("enabled", True):
+            raise HTTPException(400, f"Server '{id}' is not running")
+        logger.info(f"Auto-starting server '{id}' on demand")
+        started = await manager.start_server(id)
+        if not started:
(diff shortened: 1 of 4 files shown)


fluidmcp/cli/api/management.py
```diff
@@ -1927,13 +1927,22 @@ async def run_tool(
...
+    # Auto-start the server if it's stopped but has a valid, enabled config
+    if id not in manager.processes or manager.processes[id].poll() is not None:
+        if manager.db is None:
+            raise HTTPException(400, f"Server '{id}' is not running")
+        config = await manager.db.get_server_config(id)
+        if not config or config.get("deleted_at") or not config.get("enabled", True):
+            raise HTTPException(400, f"Server '{id}' is not running")
+        safe_id = id.replace('\n', '\\n').replace('\r', '\\r')
+        logger.info(f"Auto-starting server '{safe_id}' on demand")
+        started = await manager.start_server(id)
```
rohan-ghorpade-fluidai: @shubhendu-shekhar-fluidai pass the config to start_server as well here, as if config is missing, start_server_unlocked will check for it again in db, which will cause redunant calls to DB

fluidmcp/cli/services/package_launcher.py
```diff
@@ -357,6 +357,34 @@ def _get_io_lock(name: str) -> asyncio.Lock:
...
+        process = server_manager.processes.get(server_name)
+        if process is not None and process.poll() is None:
+            return  # already running
+
+        # No running process — check DB for a valid, enabled config
+        if server_manager.db is None:
+            raise HTTPException(404, f"Server '{server_name}' not found or not running")
+
+        config = await server_manager.db.get_server_config(server_name)
+        if not config or config.get("deleted_at"):
```
rohan-ghorpade-fluidai, line 378: @shubhendu-shekhar-fluidai Same point here as well for redundant db calls

fluidmcp/cli/services/package_launcher.py
```diff
@@ -357,6 +357,34 @@ def _get_io_lock(name: str) -> asyncio.Lock:
             _io_locks[name] = asyncio.Lock()
         return _io_locks[name]
 
+    async def _ensure_server_running(server_name: str) -> None:
```
rohan-ghorpade-fluidai: @shubhendu-shekhar-fluidai Name this function something like - auto_start_stopped_server

fluidmcp/cli/services/package_launcher.py
```diff
@@ -386,15 +414,11 @@ async def proxy_jsonrpc(
         # HTTPExceptions raised within this context are tracked as error_type="network_error"
         # via RequestTimer.__exit__ → _categorize_error() → name-based matching
         with RequestTimer(collector, method):
-            # Check if server exists
-            if server_name not in server_manager.processes:
-                raise HTTPException(404, f"Server '{server_name}' not found or not running")
```
rohan-ghorpade-fluidai, line 391: @shubhendu-shekhar-fluidai Also check if the process is alive or not before calling the auto_restart_server function, as function calls are expensive. And do this for every location where you are calling this new function

rohan-ghorpade-fluidai (changes requested):

fluidmcp/cli/services/package_launcher.py
shubhendu-shekhar-fluidai: Done, renamed `_ensure_server_running ` to `auto_start_stopped_server` across all call sites.

fluidmcp/cli/services/package_launcher.py
shubhendu-shekhar-fluidai: Added a fast-path check at the top of `auto_start_stopped_server` — if the process exists and poll() is None (alive), we return immediately without touching the DB. Applies to all 4 endpoints in package_launcher.py and the `run_tool` endpoint in management.py.

fluidmcp/cli/api/management.py
shubhendu-shekhar-fluidai: Fixed: the fetched config is now passed directly to` start_server(id, config=config)`, so `_start_server_unlocked` skips the second DB fetch it would otherwise make when config is None.

abhinav-fluidai (approved):
Readiness and security review passed. Prior requested changes are resolved in the current diff; no large-PR split is needed. Security blast radius: HIGH due shared MCP proxy/lifecycle request handling. No actionable security findings.

abhinav-fluidai (approved):

PR #761, opened 2026-05-06
MCP-574: fix: Routing Issues, UI deep-link routing behind reverse proxies

shubhendu-shekhar-fluidai:
## Summary

- **`BrowserRouter basename=\"/ui\"`** — React Router was matching routes as `/status` while the app is served at `/ui/`, so `/ui/status` never matched any route. Adding `basename` fixes all deep links.
- **Replaced `StaticFiles` mount with explicit route handlers** — `StaticFiles(html=True)` falls back to `404.html` (not `index.html`) for unknown sub-paths, breaking direct navigation to `/ui/status` etc. New handlers serve `index.html` as the SPA fallback for any unmatched `/ui/*` path.
- **Eliminated absolute-URL redirect** — Starlette redirects `/ui` → `/ui/` using the raw `Host` header, which resolves to `localhost:port` behind GitHub Codespaces. Serving `index.html` directly from `GET /ui` removes the redirect entirely.
- **Bare route redirects** — `/status`, `/servers`, `/inspector` etc. now 302 to `/ui/<route>` so copy-pasted links without the `/ui` prefix work.
- **`proxy_headers=True` on uvicorn** — trusts `X-Forwarded-Host`/`X-Forwarded-Proto` from Railway/Codespaces for correct URL construction elsewhere (OpenAPI docs, etc.).

## Test plan

- [x] `/ui` loads landing page (no redirect to localhost)
- [x] `/ui/` loads landing page
- [x] `/ui/status` loads status page directly (deep link)
- [x] `/ui/servers` loads servers page directly
- [x] `/ui/inspector` loads inspector page directly
- [x] `/status` redirects to `/ui/status` and loads correctly
- [x] `/inspector` redirects to `/ui/inspector` and loads correctly
- [x] `/ui/nonexistent` falls back to landing page (SPA catch-all)
- [x] Static assets (`/ui/assets/*.js`, `*.css`) still served correctly

### DIFF
--- a/fluidmcp/cli/services/frontend_utils.py
+++ b/fluidmcp/cli/services/frontend_utils.py
@@ -10,7 +10,8 @@ This module provides functionality to:
 from pathlib import Path
 from typing import Optional
 from loguru import logger
-from fastapi import FastAPI
+from fastapi import FastAPI, Request
+from fastapi.responses import FileResponse, RedirectResponse
 from fastapi.staticfiles import StaticFiles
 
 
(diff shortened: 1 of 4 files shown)


shivamFluidAi (approved):
looks fine

fluidmcp/cli/services/frontend_utils.py
```diff
@@ -135,26 +135,54 @@ def setup_frontend_routes(
...
+        async def ui_root():
+            return FileResponse(str(index_html))
+
+        # SPA catch-all: any /ui/{path} that isn't a real file gets index.html.
+        # Starlette's StaticFiles(html=True) falls back to 404.html (not index.html)
+        # when a path doesn't match a file, so deep-links like /ui/status would 404.
+        @app.get("/ui/{path:path}", include_in_schema=False)
+        async def ui_spa(path: str):
+            candidate = frontend_dist / path
+            if candidate.exists() and candidate.is_file():
```
abhinav-fluidai, line 185: This constructs a filesystem path directly from the user-controlled catch-all route. Requests containing dot segments or encoded traversal can make `frontend_dist / path` resolve outside the built UI directory, and the following `FileResponse` would serve any readable file that exists. Please resolve the candidate path and ensure it remains under `frontend_dist` before serving it, or use Starlette's `StaticFiles` path lookup/safe-join behavior for real files and fall back to `index.html` otherwise.

abhinav-fluidai (changes requested):
Readiness gate passed: the PR is mergeable, not draft, focused, and prior discussion shows no unresolved readiness feedback. Security review found one blocking issue in the new SPA fallback: the catch-all static file handler can resolve user-controlled paths outside the frontend dist directory. Please add a containment check or delegate file serving to Starlette StaticFiles/safe path handling before merge.

fluidmcp/cli/services/frontend_utils.py
shubhendu-shekhar-fluidai: Verified the issue SPA catch-all handler constructed file paths directly from user input without validation. I have added
- path resolution: Convert both candidate path and frontend_dist to absolute paths. 
- Fail closed: Traversal attempts serve index.html (no error messages that leak filesystem info)
- Added containment check: Use Path.relative_to() to verify resolved path stays within frontend_dist
- No information leakage: Same response for traversal attempts as legitimate missing files

abhinav-fluidai (approved):
Readiness gate passed: the PR is mergeable, not draft, and focused on UI routing/deep-link handling. The prior SPA fallback path traversal concern is addressed by resolving the candidate path and enforcing containment under the frontend dist directory before serving files. Security review found no actionable blockers.

abhinav-fluidai (approved):

PR #766, opened 2026-05-07
[MCP-578] : TRUNK-1 : Auto port allocation for HTTP-transport MCP servers

rohan-ghorpade-fluidai:
Type : New Feature

### DIFF
--- a/docs/specs/streamable_http_transport.md
+++ b/docs/specs/streamable_http_transport.md
@@ -1,0 +1,577 @@
+# Spec: Streamable-HTTP Transport for MCP Servers
+
+**Branch:** `http_streamable_spec`
+**Status:** Draft
+
+---
+
+## 1. Context and Problem Statement
+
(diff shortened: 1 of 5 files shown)


shubhendu-shekhar-fluidai (approved):
PR looks good to me. The core idea is sound, and the SSE path looks correct. 
**Tests can be added.**

**Please check these blockers before merging**:
-  `_handshake_http_subprocess` always returns None — but ports aren't released. The http transport stub should either be removed or produce a clear, user-facing error (not a silent warning + None).
- Stale entries in `_allocated_ports` on server crash. `_cleanup_server` (server_manager.py:1170-1182) releases ports, but only if `_cleanup_server` is called. If the process monitor (`_check_server`) calls `_cleanup_server`, that path looks correct — but on asyncio.TimeoutError during startup (server_manager.py:199-214, `_release_port` is never called for the `allocated_port` that was injected before spawn. The `allocated_port` would be leaked.

PR #775, opened 2026-05-08
[MCP-583] : TRUNK-3 : Add streamable-http handshake to ServerManager

rohan-ghorpade-fluidai:
Type : New Feature
Description : Handshake to start http server (initialize)

PR #777, opened 2026-05-08
[MCP-584] : TRUNK-4 : Add tool discovery for streamable-http servers

rohan-ghorpade-fluidai:
Type : New Feature
Description : Generalized Tool Discovery for SSE and HTTP Servers

### DIFF
--- a/fluidmcp/cli/services/server_manager.py
+++ b/fluidmcp/cli/services/server_manager.py
@@ -1095,7 +1095,7 @@ class ServerManager:
             return None
 
         # Discover and cache tools via HTTP
-        await self._discover_and_cache_tools_sse(id, url)
+        await self._discover_and_cache_tools_network(id, url, endpoint_path="/messages/")
 
         return NetworkSubprocessHandle(process=process, base_url=url, transport="sse")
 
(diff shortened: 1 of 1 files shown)


fluidmcp/cli/services/server_manager.py
```diff
@@ -1176,55 +1176,32 @@ async def _handshake_http_subprocess(
             return None
 
         # Discover and cache tools via POST /mcp
-        tools_request = {"jsonrpc": "2.0", "id": 1, "method": "tools/list"}
-        try:
-            async with httpx.AsyncClient(timeout=10.0) as client:
-                resp = await client.post(mcp_url, json=tools_request)
-                resp.raise_for_status()
```
rohan-ghorpade-fluidai, line 1210: The same code exists in `_discover_and_cache_tools_network`, hence removed from there as it was duplicate code

fluidmcp/cli/services/server_manager.py
```diff
@@ -1203,57 +1203,44 @@ async def _handshake_http_subprocess(
...
-        except Exception as e:
-            logger.warning(f"[HTTP] Tool discovery failed for '{id}': {e}")
+        await self._discover_and_cache_tools_network(id, base_url, session_id=session_id)
 
         return NetworkSubprocessHandle(process=process, base_url=base_url, transport="http", session_id=session_id)
 
-    async def _discover_and_cache_tools_sse(self, server_id: str, base_url: str) -> None:
+    async def _discover_and_cache_tools_network(
+        self, server_id: str, base_url: str, session_id: str = None
+    ) -> None:
```
shubhendu-shekhar-fluidai, line 1212: session_id: str = None should be session_id: Optional[str] = None — Optional is already imported at line 14.

shubhendu-shekhar-fluidai (approved):
PR looks good. This directly addresses the duplication concern from the PR #775

**Also take a look at these concerns claude flagged:**

1. session_id=None sentinel — line 1226 in server_manager.py
        `if session_id is not None:`
The method branches on `session_id is not None` to decide URL, headers, and response parsing — but an HTTP server that doesn't issue a session ID would be silently treated as SSE (wrong URL` /messages/`, no `Accept` header). Transport type should be passed explicitly as a `transport: str` parameter so the branching is unambiguous and independent of session state.

2. SSE path skips _parse_mcp_response — line 1243 in server_manager.py
               ` response = _parse_mcp_response(resp) if session_id is not None else resp.json()`
`_parse_mcp_response` already handles both plain JSON and SSE-framed responses safely. Using `resp.json()` for the SSE path is inconsistent and will silently break if any SSE server returns SSE-framed data on `/messages/`. Just use _parse_mcp_response(resp) for both branches.

abhinav-fluidai (approved):

PR #782, opened 2026-05-08
[MCP-553] : BugFix InMemory Get Env

rohan-ghorpade-fluidai:
Type : BugFix

### DIFF
--- a/fluidmcp/cli/repositories/database.py
+++ b/fluidmcp/cli/repositories/database.py
@@ -761,6 +761,7 @@ class DatabaseManager(PersistenceBackend):
                     filtered_env = {
                         k: v for k, v in env.items()
                         if v
+                        and isinstance(v, str)
                         and v.strip()
                         and not v.strip().startswith("${")
                         and not is_placeholder(v)
--- a/fluidmcp/cli/repositories/database.py
+++ b/fluidmcp/cli/repositories/database.py
@@ -760,7 +760,10 @@ class DatabaseManager(PersistenceBackend):
                     # Filter out placeholder and empty values
                     filtered_env = {
                         k: v for k, v in env.items()
-                        if v and v.strip() and not is_placeholder(v)
+                        if v
+                        and v.strip()
+                        and not v.strip().startswith("${")
+                        and not is_placeholder(v)
                     }
                     return filtered_env if filtered_env else None
                 return None
--- a/fluidmcp/cli/repositories/memory.py
+++ b/fluidmcp/cli/repositories/memory.py
@@ -10,6 +10,7 @@ from collections import deque, defaultdict
 from datetime import datetime
 from loguru import logger
 from .base import PersistenceBackend, DuplicateKeyError
+from ..utils.env_utils import is_placeholder
 
 
 class InMemoryBackend(PersistenceBackend):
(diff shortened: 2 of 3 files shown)


shubhendu-shekhar-fluidai (approved):
The pr is good. Implements get_instance_env in InMemoryBackend, the method was missing from the in-memory repository and was causing bugs. NotImplementedError (or similar) 
Adds an extra ${...} prefix check in the database.py filtering logic.

Some Issues: 
- Merge Conflicts.
- Inconsistency: database.py is missing the isinstance(v, str) guard.
- No error handling in InMemoryBackend.get_instance_env — Unlike the database version which wraps the call in try/except, the memory version has no error handling.

abhinav-fluidai (approved):

fluidmcp/cli/repositories/database.py
```diff
@@ -760,7 +760,10 @@ async def get_instance_env(self, server_id: str) -> Optional[Dict[str, str]]:
                     # Filter out placeholder and empty values
                     filtered_env = {
                         k: v for k, v in env.items()
-                        if v and v.strip() and not is_placeholder(v)
+                        if v
+                        and v.strip()
```
abhinav-fluidai, line 765: This still evaluates `v.strip()` before checking `isinstance(v, str)`. A truthy non-string env value will raise AttributeError and the database implementation catches it by returning None, dropping the whole env result. Please add the same `isinstance(v, str)` guard used in memory.py before any string methods.

fluidmcp/cli/repositories/memory.py
```diff
@@ -118,6 +119,35 @@ async def reset_instance_state(self, server_id: str) -> bool:
...
+        and whitespace-only strings so callers receive only real configured values.
+
+        Args:
+            server_id: Server identifier
+
+        Returns:
+            Dict of real environment variables, or None if the instance does
+            not exist or has no env block (or all values were placeholders).
+        """
+        instance = await self.get_instance_state(server_id)
```
abhinav-fluidai: This new in-memory implementation still has no try/except, unlike database.py. Any malformed stored env value or `get_instance_state` error will propagate from this backend instead of returning None consistently. Please wrap this method with the same error handling/logging pattern as database.py.

abhinav-fluidai (changes requested):
Readiness gate failed. GitHub now reports this PR as MERGEABLE, so the earlier merge-conflict blocker appears resolved, and the security review found no actionable security issues. The two consistency gaps called out in review/Linear are still present in the current diff: database.py still calls string methods before an isinstance(v, str) guard, and memory.py still lacks the try/except behavior used in database.py.

fluidmcp/cli/repositories/database.py
rohan-ghorpade-fluidai: done

fluidmcp/cli/repositories/memory.py
rohan-ghorpade-fluidai: done

abhinav-fluidai (approved):
Workflow review approved. GitHub reports the PR as mergeable, the PR is not a draft, and the current diff resolves the previously requested get_instance_env consistency fixes: database.py now guards isinstance(v, str) before string methods, and memory.py now uses try/except logging with return None. Security review found no actionable issues; blast radius is LOW because the change is limited to repository env filtering helpers.

abhinav-fluidai (approved):

PR #815, opened 2026-06-09
[MCP-618] : On-Premise Config Mode: Example Servers & Demo Config for Local Replication

rohan-ghorpade-fluidai:
Type : New Feature

### DIFF
--- a/examples/weather-mcp/server.py
+++ b/examples/weather-mcp/server.py
@@ -1,0 +1,49 @@
+"""
+Weather MCP Server
+Returns dummy weather data for a given city. No external API required.
+Runs over HTTP Streamable transport.
+"""
+
+from dotenv import load_dotenv
+from fastmcp import FastMCP
+import os
+
+load_dotenv()
+
+MCP_PORT = int(os.getenv("MCP_PORT"))
+TRANSPORT_TYPE = os.getenv("TRANSPORT_TYPE", "http")
+
+mcp = FastMCP("Weather-MCP")
+
+_DUMMY_DATA = {
+    "london":    {"temp_c": 12, "condition": "Cloudy",  "humidity_pct": 75},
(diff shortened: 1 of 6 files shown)


examples/weather-mcp/server.py
```diff
@@ -0,0 +1,49 @@
...
+Runs over HTTP Streamable transport.
+"""
+
+from dotenv import load_dotenv
+from fastmcp import FastMCP
+import os
+
+load_dotenv()
+
+MCP_PORT = int(os.getenv("MCP_PORT"))
```
shubhendu-shekhar-fluidai, line 13: Both server.py files do:
`MCP_PORT = int(os.getenv("MCP_PORT"))`

The metadata.json and demo config set "MCP_PORT": "" (empty string). int("") raises a ValueError at import time.
 The server might crashe before handling any request.

rohan-ghorpade-fluidai: @shubhendu-shekhar-fluidai Before running any server, our FluidMCP automatically injects an available port. The error will only appear if the env is not present in the metadata.json, and in that case the server should fail loudly so that env can be added, otherwise it will create conflict with other servers

shubhendu-shekhar-fluidai: Thank you for the context. Just check if the injection is conditional or happens every time.
I am approving the PR.

shubhendu-shekhar-fluidai (approved):

abhinav-fluidai (approved):
AI readiness and security review approved. Readiness gate passed: PR is mergeable, not draft, GitHub review decision is approved, and the prior MCP_PORT thread was reconciled against current config resolution/spawn code. Security review found no actionable findings; blast radius LOW.

abhinav-fluidai (approved):

PR #828, opened 2026-06-10
[MCP-626] : Add MCP_ROOT_PATH env var support for route prefix

rohan-ghorpade-fluidai:
Type : New Feature

### DIFF
--- a/fluidmcp/cli/cli.py
+++ b/fluidmcp/cli/cli.py
@@ -573,7 +573,7 @@ def github_command(args, secure_mode: bool = False, token: str = None) -> None:
 
             # Mount unified dynamic router (same as serve)
             mcp_router = create_dynamic_router(server_manager)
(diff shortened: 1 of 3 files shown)


shubhendu-shekhar-fluidai (approved):
Looks Good to me. Some suggestions:

1. FastAPI requires router prefixes to start with / if non-empty. If someone sets MCP_ROOT_PATH=mcp (without leading slash), FastAPI raises a ValueError at startup with a cryptic error. Should we add a guard?
2. No test verifications. Given how easy it is to misconfigure FastAPI prefix routing, even a single smoke test would be valuable.

abhinav-fluidai (approved):
AI readiness/security review: no unresolved substantive readiness blockers found. Security review approved; no realistic exploit path identified in the MCP_ROOT_PATH router-prefix changes.

abhinav-fluidai (approved):

PR #829, opened 2026-06-10
[MCP-627] : Add env_file support for per-server environment variables

rohan-ghorpade-fluidai:
Type : New Feature

### DIFF
--- a/fluidmcp/cli/services/server_manager.py
+++ b/fluidmcp/cli/services/server_manager.py
@@ -768,7 +768,24 @@ class ServerManager:
             args = config.get("args", [])
             env_vars = config.get("env", {})
 
-            # Load env_file if specified, merging before inline env (inline takes precedence)
+            # --- env_file support ---
+            # Allows secrets/env vars to be stored in a .env file on disk rather than
+            # inlined in the config JSON. This keeps production configs clean and avoids
+            # committing secrets. Usage in config:
+            #
+            #   "env_file": "servers/myserver/.env"   (relative to working_dir)
+            #   "env_file": "/absolute/path/.env"
+            #
+            # See examples/env-file-config.json for the full config structure supported.
+            #
+            # Rules:
+            #   - Relative paths are resolved relative to working_dir.
+            #   - The file must reside under install_path or working_dir (prevents
+            #     path-traversal attacks reading files outside the server tree).
+            #   - Inline "env" keys overlay on top of file values, so per-run overrides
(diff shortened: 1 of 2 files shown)


fluidmcp/cli/services/server_manager.py
```diff
@@ -767,6 +767,80 @@ async def _spawn_mcp_process(self, id: str, config: Dict[str, Any]) -> Optional[
...
+            #
+            # Rules:
+            #   - Relative paths are resolved relative to working_dir.
+            #   - The file must reside under install_path or working_dir (prevents
+            #     path-traversal attacks reading files outside the server tree).
+            #   - Inline "env" keys overlay on top of file values, so per-run overrides
+            #     still work without modifying the file.
+            #   - Standard .env syntax: KEY=VALUE, # comments, blank lines ignored,
+            #     optional surrounding single/double quotes stripped from values.
+            env_file_path = config.pop("env_file", None)
```
shubhendu-shekhar-fluidai: The config dict is popped in-place. If `_spawn_mcp_process` is ever called more than once for the same config object (e.g. restart), the `env_file` key will be gone on the second call. This is a latent bug. Should use `config.get(...`) and handle separately, or work on a copy.

fluidmcp/cli/services/server_manager.py
```diff
@@ -767,6 +767,80 @@ async def _spawn_mcp_process(self, id: str, config: Dict[str, Any]) -> Optional[
...
+            if env_file_path:
+                early_working_dir = config.get("working_dir", ".")
+                early_install_path = config.get("install_path", ".")
```
shubhendu-shekhar-fluidai, line 792: When neither` install_path` nor `working_dir` is set, the guard resolves both to the process CWD. That means `env_file: "../../etc/passwd"` would pass the security check if the CWD is /, or any file within the resolved CWD tree would be allowed. Should `install_path` be required explicitly set before allowing `env_file`.

shubhendu-shekhar-fluidai (changes requested):

fluidmcp/cli/services/server_manager.py
rohan-ghorpade-fluidai: resolved

fluidmcp/cli/services/server_manager.py
rohan-ghorpade-fluidai: enforcing install_path would break direct server configs that never set it, so not required

PR #847, opened 2026-06-16
MCP-276 : feat: Full observability stack — OTEL traces + Loki logs + Prometheus metrics + Grafana

dhanesh-gore-fluidai:
## Summary

- **OpenTelemetry tracing** — `otel.py` initialises a `TracerProvider` with OTLP HTTP export to Grafana Tempo; `context.py` stores `trace_id`/`span_id` in per-request `ContextVar`s; `TraceContextMiddleware` in `server.py` injects them on every request
- **Structured JSON logging** — `cli.py` now emits one JSON object per log line (`timestamp`, `level`, `message`, `trace_id`, `span_id`) when `LOG_FORMAT=json` (default in containers); Promtail parses and ships these to Loki
- **Full Docker Compose stack** — `docker-compose.observability.yml` wires FluidMCP → Prometheus → Tempo → Loki → Promtail → Grafana in a single `docker compose up`
- **Pre-wired Grafana** — all three datasources provisioned automatically; cross-signal correlation enabled (logs ↔ traces ↔ metrics exemplars)

## Services

| Service | Port | Purpose |
|---|---|---|
| FluidMCP | 8099 | Gateway (emits OTLP traces + JSON logs + `/metrics`) |
| Prometheus | 9090 | Scrapes `/metrics` every 10 s, holds alert rules |
| Grafana Tempo | 3200 / 4318 | Receives OTLP HTTP traces from FluidMCP |
| Grafana Loki | 3100 | Receives logs from Promtail |
| Promtail | — | Tails Docker container logs → Loki |
| Grafana | 3000 | Unified UI (admin/admin) |

## Files changed

```
requirements.txt                          ← 6 OTEL packages added
fluidmcp/cli/otel.py                      ← new: OTEL init/shutdown
fluidmcp/cli/context.py                   ← new: per-request trace context
fluidmcp/cli/server.py                    ← TraceContextMiddleware + init/shutdown hooks
fluidmcp/cli/cli.py                       ← structured JSON log sink
docker-compose.observability.yml          ← full stack compose
observability/prometheus/prometheus.yml   ← scrape config
observability/prometheus/alerts.yml       ← 5 alert rules
observability/tempo/tempo.yml             ← OTLP HTTP receiver
observability/loki/loki.yml               ← log storage, 7d retention
observability/promtail/promtail.yml       ← Docker discovery + JSON parsing
observability/grafana/provisioning/       ← datasources + dashboard provider
```

## How to run

```bash
docker compose -f docker-compose.observability.yml up --build
# Grafana → http://localhost:3000  (admin/admin)
```

## Test plan

- [ ] `docker compose up` starts all 6 services without errors
- [ ] `curl http://localhost:8099/health` returns `{"status":"healthy"}`
- [ ] `curl http://localhost:8099/metrics` returns Prometheus text format
- [ ] Grafana Explore → Loki → `{service="fluidmcp"}` shows FluidMCP log lines with `level` labels
- [ ] Grafana Explore → Tempo → search by service name returns HTTP spans
- [ ] Clicking a Loki log line with `trace_id` opens the matching trace in Tempo
- [ ] Prometheus → Status → Targets shows `fluidmcp` as UP

### DIFF
--- a/docker-compose.observability.yml
+++ b/docker-compose.observability.yml
@@ -3,22 +3,21 @@ version: "3.9"
 # FluidMCP Full Observability Stack
 # ─────────────────────────────────
 # Signals collected:
-#   Logs    → Promtail scrapes FluidMCP stderr (JSON) → Loki
+#   Logs    → Promtail tails Docker json-file logs → Loki
 #   Metrics → Prometheus scrapes /metrics every 10 s
 #   Traces  → FluidMCP OTLP HTTP → Tempo
 #   UI      → Grafana (all three datasources pre-wired)
 #
 # Usage:
-#   docker compose -f docker-compose.observability.yml up --build
+#   cp .env.observability.example .env.observability
+#   # Edit .env.observability and set GF_SECURITY_ADMIN_PASSWORD
+#   docker compose -f docker-compose.observability.yml --env-file .env.observability up --build
 #
-# URLs:
+# URLs (all bound to 127.0.0.1 — localhost only):
 #   FluidMCP   http://localhost:8099
-#   Grafana    http://localhost:3000   (admin / admin)
+#   Grafana    http://localhost:3000   (credentials from .env.observability)
 #   Prometheus http://localhost:9090
 #   Tempo      http://localhost:3200   (API / query)
-#
-# Environment:
-#   Copy .env.observability.example → .env and fill in secrets before first run.
 
 services:
 
@@ -33,24 +32,21 @@ services:
     environment:
       PORT: "8099"
       PYTHONUNBUFFERED: "1"
-      LOG_FORMAT: "json"                          # emit JSON lines for Promtail
+      LOG_FORMAT: "json"
       OTEL_ENABLED: "true"
       OTEL_SERVICE_NAME: "fluidmcp"
       OTEL_SERVICE_VERSION: "2.0.0"
-      OTEL_EXPORTER: "jaeger"                     # sends to Tempo via OTLP HTTP
+      OTEL_EXPORTER: "jaeger"
       OTEL_EXPORTER_OTLP_ENDPOINT: "http://tempo:4318/v1/traces"
-      # Uncomment and set a stable bearer token to avoid regeneration on restart:
       # FMCP_BEARER_TOKEN: "${FMCP_BEARER_TOKEN}"
-      # MongoDB (optional — omit to run without persistence):
       # MONGODB_URI: "mongodb://mongo:27017"
-    volumes:
-      # Named volume so Promtail can read the log file written by the container
-      - fluidmcp_logs:/var/log/fluidmcp
     logging:
       driver: "json-file"
       options:
         max-size: "50m"
         max-file: "5"
+        # tag is embedded in each log line's attrs field;
+        # Promtail reads it to identify fluidmcp logs without the Docker socket.
         tag: "fluidmcp"
     healthcheck:
       test: ["CMD", "curl", "-f", "http://localhost:8099/health"]
@@ -64,11 +60,14 @@ services:
       - tempo
 
   # ── Prometheus ────────────────────────────────────────────────────────────
+  # Bound to 127.0.0.1: only accessible from the local machine.
+  # --web.enable-lifecycle is intentionally omitted (unauthenticated reload/quit).
+  # --web.enable-remote-write-receiver is omitted; Tempo uses its own storage.
   prometheus:
     image: prom/prometheus:v2.51.0
     container_name: prometheus
     ports:
-      - "9090:9090"
+      - "127.0.0.1:9090:9090"
     volumes:
       - ./observability/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml:ro
       - ./observability/prometheus/alerts.yml:/etc/prometheus/alerts.yml:ro
@@ -79,20 +78,19 @@ services:
       - "--storage.tsdb.retention.time=15d"
       - "--web.console.libraries=/usr/share/prometheus/console_libraries"
       - "--web.console.templates=/usr/share/prometheus/consoles"
-      - "--web.enable-lifecycle"
-      # Enable remote-write receiver so Tempo can push exemplars back (optional)
-      - "--web.enable-remote-write-receiver"
     networks:
       - observability
 
   # ── Grafana Tempo (distributed tracing backend) ───────────────────────────
+  # 4318 (OTLP HTTP) bound to 127.0.0.1; FluidMCP reaches it via the internal
+  # Docker network (service name "tempo"), not through the host port.
   tempo:
     image: grafana/tempo:2.4.1
     container_name: tempo
     ports:
-      - "3200:3200"    # Tempo HTTP API / Grafana query
-      - "4317:4317"    # OTLP gRPC (optional, for gRPC exporters)
-      - "4318:4318"    # OTLP HTTP (FluidMCP sends here)
+      - "127.0.0.1:3200:3200"
+      - "127.0.0.1:4317:4317"
+      - "127.0.0.1:4318:4318"
     volumes:
       - ./observability/tempo/tempo.yml:/etc/tempo/tempo.yml:ro
       - tempo_data:/var/tempo
@@ -105,7 +103,7 @@ services:
     image: grafana/loki:2.9.4
     container_name: loki
     ports:
-      - "3100:3100"
+      - "127.0.0.1:3100:3100"
     volumes:
       - ./observability/loki/loki.yml:/etc/loki/loki.yml:ro
       - loki_data:/loki
@@ -113,15 +111,17 @@ services:
     networks:
       - observability
 
-  # ── Promtail (log shipper: Docker logs → Loki) ────────────────────────────
+  # ── Promtail (log shipper: Docker json-file logs → Loki) ──────────────────
+  # No Docker socket mount. Promtail reads log files directly from
+  # /var/lib/docker/containers (already :ro) using a glob path.
+  # The container "tag" attribute (set in fluidmcp logging.options) is extracted
+  # from each log line's JSON envelope to identify the fluidmcp container.
   promtail:
     image: grafana/promtail:2.9.4
     container_name: promtail
     volumes:
       - ./observability/promtail/promtail.yml:/etc/promtail/promtail.yml:ro
-      # Mount Docker socket so Promtail can discover and tail container logs
       - /var/lib/docker/containers:/var/lib/docker/containers:ro
-      - /var/run/docker.sock:/var/run/docker.sock
     command: ["-config.file=/etc/promtail/promtail.yml"]
     depends_on:
       - loki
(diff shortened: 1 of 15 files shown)


docker-compose.observability.yml
```diff
@@ -0,0 +1,164 @@
...
+
+  # ── Promtail (log shipper: Docker logs → Loki) ────────────────────────────
+  promtail:
+    image: grafana/promtail:2.9.4
+    container_name: promtail
+    volumes:
+      - ./observability/promtail/promtail.yml:/etc/promtail/promtail.yml:ro
+      # Mount Docker socket so Promtail can discover and tail container logs
+      - /var/lib/docker/containers:/var/lib/docker/containers:ro
+      - /var/run/docker.sock:/var/run/docker.sock
```
abhinav-fluidai: Mounting `/var/run/docker.sock` gives the Promtail container Docker daemon access, which is effectively host-level control if Promtail or its image is compromised. Avoid Docker socket discovery here; tail the mounted container log path with static/file targets, or put discovery behind a tightly restricted Docker socket proxy.

docker-compose.observability.yml
```diff
@@ -0,0 +1,164 @@
...
+
+  # ── Grafana (visualization) ────────────────────────────────────────────────
+  grafana:
+    image: grafana/grafana:10.4.2
+    container_name: grafana
+    ports:
+      - "3000:3000"
+    environment:
+      GF_SECURITY_ADMIN_USER: admin
+      GF_SECURITY_ADMIN_PASSWORD: admin
```
abhinav-fluidai: This hardcodes Grafana `admin/admin` while publishing `3000:3000`. On Docker hosts where the port is reachable beyond localhost, anyone can log in and query the provisioned Loki/Tempo/Prometheus data, including request logs and traces. Bind the port to localhost for local-only use and require `GF_SECURITY_ADMIN_PASSWORD` from env/secrets.

docker-compose.observability.yml
```diff
@@ -0,0 +1,164 @@
...
+      - ./observability/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml:ro
+      - ./observability/prometheus/alerts.yml:/etc/prometheus/alerts.yml:ro
+      - prometheus_data:/prometheus
+    command:
+      - "--config.file=/etc/prometheus/prometheus.yml"
+      - "--storage.tsdb.path=/prometheus"
+      - "--storage.tsdb.retention.time=15d"
+      - "--web.console.libraries=/usr/share/prometheus/console_libraries"
+      - "--web.console.templates=/usr/share/prometheus/consoles"
+      - "--web.enable-lifecycle"
```
abhinav-fluidai: `--web.enable-lifecycle` is enabled while Prometheus is published on `9090:9090` without authentication. A network-reachable attacker can hit lifecycle endpoints and disrupt monitoring. Keep lifecycle disabled by default, or bind Prometheus to localhost and protect it behind auth.

abhinav-fluidai (changes requested):
Readiness gate passed for PR #847: the PR is open, not draft, mergeable, cohesive around the observability stack, and there are no unresolved PR review threads or substantive PR #847 discussion items. Security review blocks merge because the new Compose stack publishes observability services with unsafe defaults: host Docker socket access for Promtail, public Prometheus lifecycle endpoints, and hardcoded Grafana admin credentials on a published port. Split recommendation: not needed.

dhanesh-gore-fluidai:
## Security fixes applied — ready for re-review

All three issues raised by the AI readiness gate have been addressed in commit `b7812e3`.

### 1. Promtail Docker socket exposure — **Fixed**
- Removed `/var/run/docker.sock` volume mount entirely from Promtail
- Switched to a static file glob on `/var/lib/docker/containers/*/*-json.log` (already mounted `:ro`)
- Container identity is now derived from the `tag` field injected by Docker's `json-file` driver (configured via `logging.options.tag: fluidmcp` in the compose file) — no socket access required

### 2. Hardcoded Grafana admin credentials — **Fixed**
- `GF_SECURITY_ADMIN_PASSWORD` is now `${GF_SECURITY_ADMIN_PASSWORD}` with **no default** — compose fails fast at startup if the variable is unset
- Added `GF_AUTH_ANONYMOUS_ENABLED: "false"` explicitly
- Grafana port bound to `127.0.0.1:3000` (localhost only, not `0.0.0.0`)
- Added `.env.observability.example` with setup instructions; `.gitignore` updated to track it

### 3. Unauthenticated Prometheus admin endpoints — **Fixed**
- Removed `--web.enable-lifecycle` (allowed unauthenticated config reload / process shutdown)
- Removed `--web.enable-remote-write-receiver`
- Prometheus port bound to `127.0.0.1:9090`
- Tempo (3200, 4317, 4318) and Loki (3100) ports also bound to `127.0.0.1`

observability/promtail/promtail.yml
```diff
@@ -0,0 +1,75 @@
...
+  # We unwrap it with the "docker" pipeline stage and use the "tag" attr
+  # (set to "fluidmcp" in the compose logging.options) to identify the container.
+  - job_name: docker_files
+
+    static_configs:
+      - targets: ["localhost"]
+        labels:
+          job: docker
+          # Glob picks up every container's log file on this host
+          __path__: /var/lib/docker/containers/*/*-json.log
```
abhinav-fluidai, line 28: I considered the follow-up reply that the Docker socket was replaced with static file scraping, and that does fix host Docker daemon access. This glob still reads every Docker container log on the host, though, and the pipeline only enriches `{container="fluidmcp"}` entries; it does not drop other containers before sending them to Loki. On a shared Docker host this can leak unrelated service logs and secrets into the FluidMCP Grafana/Loki stack. Please restrict the mounted path/target to FluidMCP logs only, or add an explicit drop stage for non-FluidMCP entries before they are pushed.

abhinav-fluidai (changes requested):
Readiness gate passed: PR #847 is open, not draft, mergeable, cohesive around the observability stack, and the earlier Docker socket/Grafana credential/Prometheus lifecycle blockers are fixed in the current diff. Split recommendation: not needed. Security still blocks merge because the Promtail replacement tails the host-wide Docker container log tree and does not drop non-FluidMCP logs before shipping to Loki.

observability/promtail/promtail.yml
abhinav-fluidai, line 28: I considered the 2026-06-17 reply that static file scraping replaces the Docker socket. That fixes daemon access, but this glob still reads every Docker container log on the host and the later `match` stage only enriches `{container="fluidmcp"}` entries; it does not drop non-matching entries before they are sent to Loki. On a shared host this can leak unrelated service logs/secrets into this stack. Please restrict the mounted target to FluidMCP logs only or add an explicit drop stage for non-FluidMCP entries before push.

abhinav-fluidai (changes requested):
Readiness gate passed: PR #847 is open, non-draft, mergeable, cohesive around the observability stack, and the previous Docker socket, Grafana credential, and Prometheus lifecycle blockers are fixed in the current diff. Split recommendation: not needed. Security still blocks merge: after considering the author's 2026-06-17 reply, the current Promtail config still mounts and scrapes the host-wide Docker log tree and does not drop non-FluidMCP entries before pushing to Loki.

shubhendu-shekhar-fluidai (approved):
Looks OK to me. But these should be considered:

1. LOG_FORMAT defaults to "json" - will break every local dev environment out of the box.
log_format = os.getenv("LOG_FORMAT", "json").lower()

The default should be "text" (human-readable), and "json" should only be the default in containers (in docker-compose.observability.yml). The compose file already sets LOG_FORMAT: "json" - there's no need for this to be the global default.

2. TraceContextMiddleware reads trace IDs before FastAPI instrumentation runs

In create_app(), middleware is added in this order:

app.add_middleware(TraceContextMiddleware) — line ~194
instrument_fastapi_app(app) — line ~299 (FastAPIInstrumentor wraps ASGI at the outermost layer)
Starlette executes add_middleware() calls in reverse (last-added runs first). TraceContextMiddleware is added before FastAPIInstrumentor.instrument_app() wraps the app. The OTEL span won't exist yet when TraceContextMiddleware.dispatch() runs, so span.is_recording() will return False on every request. The middleware should be added after instrument_fastapi_app(), or the trace ID should be extracted inside the route handler using trace.get_current_span().

3. Shutdown race: _otel_shutdown_done flag checked outside the lock

with _otel_shutdown_lock:
    if _otel_shutdown_done:
        return True
if not _otel_initialized or _otel_provider is None:  # ← outside lock
    ...
Between releasing the lock and the subsequent checks, another thread could partially initialize. The entire early-exit block should be inside the lock.

observability/promtail/promtail.yml
abhinav-fluidai, line 28: I considered the follow-up reply that static file scraping replaces Docker socket access. That removes Docker daemon control, but this glob still reads every Docker container log on the host and the later match stage only enriches `{container="fluidmcp"}` entries; it does not drop non-matching entries before they are sent to Loki. On a shared Docker host this can leak unrelated service logs or secrets into the FluidMCP Grafana/Loki stack. Restrict the mounted path/target to FluidMCP logs only, or add an explicit drop stage for non-FluidMCP entries before push.

abhinav-fluidai (changes requested):
Readiness gate passed: PR #847 is open, non-draft, mergeable, and cohesive around the observability stack; split recommendation not needed. Security still blocks merge: after considering the author reply that static file scraping replaced Docker socket access, the current Promtail config still reads the host-wide Docker container log tree and does not drop non-FluidMCP entries before pushing to Loki.

dhanesh-gore-fluidai:
## All 10 review issues addressed — commit `a377a1d`

### Blocking (fixed)

**#1 — LOG_FORMAT default `json` breaks local dev**
`cli.py`: default changed to `"text"`. JSON mode only activates when `LOG_FORMAT=json` is explicitly set (already done in `docker-compose.observability.yml`). Local devs get colour output with no config change.

**#2 — TraceContextMiddleware runs before OTEL span exists**
`server.py`: `app.add_middleware(TraceContextMiddleware)` moved to *after* `instrument_fastapi_app()`. Starlette reverses `add_middleware()` order so last-added runs outermost — placing it here puts it inside the FastAPIInstrumentor ASGI wrapper. Trace/span IDs are now extracted after `call_next()` returns (when the span is guaranteed to be recording).

### Medium (fixed)

**#3 — `shutdown_otel` flag read outside the lock (TOCTOU race)**
`otel.py`: all early-exit checks (`_otel_shutdown_done`, `_otel_initialized`, `_otel_provider`) moved inside the lock. The done flag is set to `True` inside the lock before any work begins, preventing concurrent re-entry.

**#4 — `OTEL_EXPORTER: "jaeger"` misleading**
`otel.py`: default changed from `"jaeger"` to `"otlp"`; condition updated from `("jaeger","both")` to `("otlp","both")`. `docker-compose.observability.yml`: value updated to `"otlp"`.

**#5 — `opentelemetry-instrumentation-httpx` in requirements but never wired up**
`otel.py`: added `_instrument_httpx()` called from `init_otel()`. All outbound httpx calls (to Prometheus, Loki, Tempo) now appear as child spans automatically.

### Low (fixed)

**#6 — Blocking `httpx.get()` inside async startup**
`otel.py`: endpoint verification moved to a daemon `threading.Thread` (fire-and-forget). Startup no longer blocks on the 2-second timeout.

**#7 — `disableDeletion: false` allows dashboard deletion in Grafana UI**
`grafana/provisioning/dashboards/dashboard.yml`: changed to `disableDeletion: true`.

**#8 — Prometheus scrape interval mismatch (15 s global vs 10 s job)**
`prometheus.yml`: global `scrape_interval` and `evaluation_interval` unified to `10s`; per-job override removed. `alerts.yml` `interval` aligned to `10s`.

**#9 — Loki `auth_enabled: false` has no production caveat**
`loki.yml`: added comment with link to Loki authentication docs.

**#10 — `examples/grafana-dashboard.json` not in PR diff**
Verified the file exists in the repo — no change needed.

fluidmcp/cli/server.py
```diff
@@ -28,6 +30,39 @@
...
+    """Inject OTEL trace/span IDs into per-request contextvars for log correlation.
+
+    This middleware must be added AFTER instrument_fastapi_app() so that it sits
+    inside the FastAPIInstrumentor ASGI wrapper. Starlette reverses add_middleware()
+    order, meaning the last-added middleware runs first. By adding this one after
+    FastAPIInstrumentor wraps the app, the OTEL span is already active when
+    dispatch() is called, so span.is_recording() reliably returns True.
+
+    The trace/span IDs are extracted after call_next() returns so we capture the
+    span that was active during the entire request, not just at entry.
```
abhinav-fluidai: This awaits the entire downstream request before setting trace_id/span_id, so every application log emitted inside call_next runs while both ContextVars are still empty. I considered the middleware-order rationale: it may ensure a span exists, but post-response assignment is still too late for request logs. Capture and set the active span context before awaiting call_next, then reset/clear it in a finally block so concurrent requests remain isolated.

abhinav-fluidai (changes requested):
Readiness is blocked by the trace/log correlation middleware timing. I considered the author's explanation that middleware ordering ensures an active OpenTelemetry span, but the ContextVars are populated only after call_next returns, so logs emitted during request handling cannot contain the advertised trace_id/span_id. Move trace-context extraction and assignment before awaiting downstream request handling, while preserving an active span, and clear/reset the context in a finally block. Security review found no separate realistic exploit path in the current diff; security verdict: APPROVE, blast radius: HIGH. Split recommendation: not needed.

PR #849, opened 2026-06-17
[MCP-642]-updating package manager

aditya-shukla-fluidai:


shubhendu-shekhar-fluidai (changes requested):
1. `stderr=subprocess.PIPE` hardcoded -- breaks on-disk log writing
In server_manager.py the diff changes `stderr=stderr_fh` to `stderr=subprocess.PIPE` and sets log_fh = None if `stderr_fh == subprocess.PIPE` else stderr_fh. This means:
If _open_stderr_log() succeeded (the normal path), stderr_fh is a file object, log_fh becomes that file object, and the drainer writes to it - correct.
But the subprocess is now given `stderr=subprocess.PIPE` instead of the log file handle. The log file handle (log_fh) is now an orphaned open file descriptor that nothing writes to directly. The drainer writes via Python's `log_fh.write(line)` - which works - but the file is opened in `_open_stderr_log` then never registered for cleanup in the log_fh code path.
The real problem: the crash-detection branch at line 984 still checks if `stderr_fh == subprocess.PIPE` — this is now always true since `stderr=subprocess.PIPE` is hardcoded. The branch reading self._read_crash_stderr(id) (which reads from the log file) will never be reached. All immediate-crash diagnostics fall back to `process.communicate()`, which tries to read from the PIPE that the drainer is also consuming — a race condition that can deadlock.

2. log_fh is never closed if the drainer silently swallows all errors
except (OSError, ValueError):
    pass
If `log_fh.write()` raises permanently (disk full, file deleted), the drainer silently continues, but log_fh is never explicitly closed. The `_close_stderr_log` method closes what it tracks in self._stderr_logs; if `log_fh` is separate from that tracking, it leaks a file descriptor for the lifetime of the process.

shubhendu-shekhar-fluidai (approved):
Looks Good to me. Just resolve the merge conflicts

fluidmcp/cli/api/inspector.py
```diff
@@ -0,0 +1,284 @@
...
+    host = parsed.hostname or ""
+    if not host:
+        raise HTTPException(400, "URL must include a host")
+
+    try:
+        addr = ipaddress.ip_address(host)
+        if addr.is_private or addr.is_loopback or addr.is_link_local or addr.is_reserved:
+            raise HTTPException(400, "Connections to private/internal addresses are not allowed")
+    except ValueError:
+        pass  # hostname, not a bare IP — DNS rebinding is out of scope here
```
abhinav-fluidai, line 64: This leaves the inspector SSRF protection bypassable for hostnames: a caller can provide a DNS name that resolves to loopback, link-local, or private infrastructure, and this branch skips the blocklist before `InspectorSession` sends server-side `httpx` requests. Please resolve the hostname and reject private/internal resolved addresses, and make sure the request uses the validated address or an equivalent resolver guard so DNS rebinding cannot bypass it.

fluidmcp/cli/services/inspector_session.py
```diff
@@ -0,0 +1,339 @@
...
+        raise ValueError("Only http/https URLs are allowed")
+
+    host = parsed.hostname or ""
+    if not host:
+        raise ValueError("URL must include a host")
+
+    try:
+        addr = ipaddress.ip_address(host)
+    except ValueError:
+        return  # hostname, not a bare IP — DNS rebinding is out of scope
```
abhinav-fluidai, line 29: This repeats the same hostname SSRF gap for SSE-derived endpoints: hostnames return before any DNS resolution, but the value is then used for server-side `httpx` requests. Please apply the same resolved-address private/internal network checks here as the connect endpoint, including protection against DNS rebinding between validation and request dispatch.

abhinav-fluidai (changes requested):
Readiness/security workflow review for MCP-642.

Readiness gate failed because GitHub reports this PR as CONFLICTING. The latest human review says “Looks Good to me. Just resolve the merge conflicts,” and I did not find a later GitHub Actions readiness comment that overrides the current mergeability state.

Security gate failed because the new inspector opens server-side HTTP/SSE connections to a user-supplied URL, but hostname validation does not resolve DNS before the httpx request. An authenticated caller can supply a hostname that resolves to loopback, link-local, or private infrastructure and bypass the literal-IP block. Please resolve hostnames and reject private/internal resolved addresses, and ensure the actual connection cannot re-resolve to a different internal address before request dispatch.

Required changes: resolve the merge conflicts and fix the inspector hostname/DNS SSRF validation.

PR #877, opened 2026-07-30
[MCP-667]: Fix streamable-HTTP MCP handshake: handle JSON-RPC notifications correctly

shubhendu-shekhar-fluidai:
## Summary

Fixes a bug where spec-compliant streamable-HTTP MCP clients (e.g. an OpenAI-Agents-SDK-based sandbox agent) failed to initialize against a FluidMCP-hosted MCP server, while the same server worked fine over stdio.

- `POST /{server_name}/mcp` (`proxy_jsonrpc` / `_proxy_to_http_server` in `fluidmcp/cli/services/package_launcher.py`) had no concept of JSON-RPC **notifications** (messages with no `"id"`, e.g. `notifications/initialized`).
- The upstream MCP server correctly acks a notification with an empty-body `202`, but the proxy unconditionally called `resp.json()` on every response, raising `JSONDecodeError` on that empty body — converted into an opaque `500`.
- A client doing the full 3-stage handshake (`initialize` → `notifications/initialized` → first request) never got past stage 2. A client that skips straight to stage 3 (e.g. the existing fluid-gpt engine integration) never hit this, which is why it "worked" there.

## Changes

- `_proxy_to_http_server` now detects an id-less payload and returns `(None, upstream_session_id)` without attempting to parse a response body for it.
- `proxy_jsonrpc` acks notifications with an empty `202 Accepted` instead of crashing, and propagates any `Mcp-Session-Id` header from the upstream back to the caller.
- Updated the three other call sites (`sse_stream`, `list_tools`, `call_tool`) for the new `(response, session_id)` return signature (these build their own payloads with an `id`, so behavior is unchanged for them).

**Also bundled in this PR (unrelated to the above, folded in so there's a single PR to review today):** a one-line-diff fix for a pre-existing frontend `tsc` build error (`ResourcesPanel.tsx` TS6133: `setSelectedResourceUri` declared but never read) that has been breaking the `Push Docker image to DockerHub` workflow on every push since 2026-06-29 — meaning no new image has published since 2026-06-23, independent of anything in this PR. Without this, merging the handshake fix alone would still fail to actually publish. Verified `npm run build` (`tsc -b && vite build`, the exact Docker build step) passes clean with it included.

## Test plan

- [x] New regression tests in `tests/test_streamable_http_proxy.py` (5 tests): notification ack doesn't crash, session header propagation on notification acks and regular responses, SSE-envelope unwrapping still works, session ID header sent upstream correctly.
- [x] `pytest tests/test_streamable_http_proxy.py tests/test_concurrency_limiting.py` — 30 passed.
- [x] `npx tsc -b` and `npm run build` — clean, confirming the DockerHub build will actually succeed once this merges.

### DIFF
--- a/fluidmcp/cli/services/package_launcher.py
+++ b/fluidmcp/cli/services/package_launcher.py
@@ -7,7 +7,7 @@ import asyncio
 import time
 import threading
 import httpx
-from typing import Union, Dict, Any, Iterator, AsyncIterator
+from typing import Union, Dict, Any, Iterator, AsyncIterator, Optional, Tuple
 from pathlib import Path
 from loguru import logger
 from fastapi import Request, APIRouter, Body, Depends, HTTPException
@@ -149,7 +149,7 @@ async def _proxy_to_http_server(
     timeout: float = 60.0,
     session_id: str = None,
     client: httpx.AsyncClient = None,
-) -> dict:
+) -> Tuple[Optional[dict], Optional[str]]:
     """
     Forward a JSON-RPC request to a streamable-http MCP server via POST /mcp.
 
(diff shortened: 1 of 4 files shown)


abhinav-fluidai (approved):
Security review approved. The change affects shared streamable-HTTP request handling (HIGH structural blast radius), but no actionable exploit path was found: upstream HTTP errors remain enforced, session IDs are only forwarded from upstream responses, and empty-body handling is limited to JSON-RPC notifications without an id.

abhinav-fluidai (approved):

